PR #65, opened 2021-07-08
Implemented drop-down autocomplete using EuiSuggest

sejli:
### Description
Search bar has a very simple dropdown autosuggest feature in the Observability Event Explorer.

### Check List


public/components/common/seach/queryBar.tsx
```diff
@@ -9,28 +9,83 @@
...
+    },
+    [value, prefix]
+  );
+
+  const onInputChange = useCallback((target) => {
+    setInputValue(target.value);
+    handleQueryChange(target.value);
+  }, [])
+
+  const append = <EuiButtonEmpty>PPL</EuiButtonEmpty>
```
joshuali925, line 78: Do we want this to be clickable? also you might want to adjust the size of `PPL`, default size might be larger than other components

public/components/common/seach/queryBar.tsx
```diff
@@ -9,28 +9,83 @@
...
+  {type: {iconType: 'search', color: 'tint1'} ,label: 'where'},
+  {type: {iconType: 'search', color: 'tint1'} ,label: 'head'},
+  {type: {iconType: 'search', color: 'tint1'} ,label: 'top'},
+  {type: {iconType: 'search', color: 'tint1'}, label: 'source'}
+]
+
+export function QueryBar(props: IQueryBarProps) {
+  
   const {
     query,
```
joshuali925, line 52: just curious, looks like the text field depends on `inputValue` to display, calls `handleQuerySearch` to search. what are `query`, `handleQueryChange`, `value` used for?

joshuali925 (commented):
a few questions

public/components/common/seach/queryBar.tsx
```diff
@@ -9,28 +9,83 @@
...
+  EuiSuggestItemProps
 } from '@elastic/eui';
+import useDebounce from 'react-use/lib/useDebounce'
 import {
   IQueryBarProps
 } from './search';
 import {
   RAW_QUERY
 } from '../../../common/constants/explorer';
+import { _termValuesToQuery } from '@elastic/eui/src/components/search_bar/query/ast_to_es_query_dsl';
```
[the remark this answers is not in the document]
sejli: yup, will be cleaning up the code a bit in my my next commit

public/components/common/seach/queryBar.tsx
```diff
@@ -9,28 +9,83 @@
...
   EuiFieldSearch,
   EuiFlexItem,
+  EuiSuggest,
+  EuiFlexGroup,
+  EuiSelectableTemplateSitewide,
+  EuiSelectableTemplateSitewideOption,
+  EuiSuggestItem,
+  EuiSuggestItemProps
 } from '@elastic/eui';
+import useDebounce from 'react-use/lib/useDebounce'
```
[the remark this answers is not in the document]
sejli: initially, i wanted to follow someone's idea about using Debounce but i couldn't get it to work. will be removing

public/components/common/seach/queryBar.tsx
```diff
@@ -9,28 +9,83 @@
...
     query,
     handleQueryChange,
     handleQuerySearch
   } = props;
 
+  const [value, setValue] = useState('');
+  const [inputValue, setInputValue] = useState(value);
+
+  const splittedModel = value.split(',');
+  const prefix = splittedModel[Math.max(0, splittedModel.length - 1)]
```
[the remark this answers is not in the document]
sejli: the value state was when i first started working on it so i'll take it out and replace the variables with inputValue. i tried out a lot of different methods to get the states working and ended with inputValue being the one i use

completely forgot there was a max() function in javascript/typescript haha, i'll get that changed to just max

public/components/common/seach/queryBar.tsx
sejli: nah, i was just following the [EuiSuggest documentation](https://elastic.github.io/eui/#/forms/suggest). it's unnecessary since the placeholder indicates that it should be PPL, just thought it would look clearer at the time. will be taking it out

public/components/common/seach/queryBar.tsx
sejli: the function that renders the search bar is `QueryBar` that takes in the `IQueryBarProps` props, which is declared in [search.tsx](https://github.com/sejli/trace-analytics/blob/feature/autocomplete/public/components/common/seach/search.tsx) with an interface creating `query`, `handleQueryChange`, and `handleQuerySearch`. `query` is the actual query that is sent to backend, which is updated by `handleQueryChange` each time the input changes (see `onInputChange` function). as mentioned earlier, `value` is redundant and i will be using `inputValue` to keep track of the string that the user types in. since `query` isn't a string that can be split into a list, i'm using `inputValue` to filter the suggestions.

public/components/common/seach/queryBar.tsx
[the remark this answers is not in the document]
joshuali925: i see, you could try lodash debounce if you need to. example here: https://github.com/opensearch-project/sql/blob/main/workbench/public/components/Main/main.tsx#L241-L242

But i don't think it's necessary to debounce since search is triggered by the enter key, not incremental

public/components/common/seach/queryBar.tsx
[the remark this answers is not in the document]
joshuali925: sorry the `max` was just me saving a few keystrokes.. my point is that the `Math.max` doesn't seem to make a difference?

public/components/common/seach/queryBar.tsx
joshuali925: got it. still a bit confused because if you keep `query` up to date using `handleQueryChange` each time the input changes, then why not use `query` instead of `inputValue`?

sejli: i split the user input using `.split(' ')`, but for some reason, it still doesn't work with `query` even though i made sure that query is a string. couldn't get it to work so i just made another `inputValue` state

public/components/common/seach/queryBar.tsx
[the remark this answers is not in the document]
sejli: actually, just tried it and it doesn't work. guess we need to use `Math.max` since it's telling me that the `max` function doesn't exist

public/components/common/seach/queryBar.tsx
```diff
@@ -9,28 +9,137 @@
...
     query,
     handleQueryChange,
     handleQuerySearch
   } = props;
 
+  const [inputValue, setInputValue] = useState('');
+
+  const splittedModel = inputValue.split(' ');
+  const prefix = splittedModel[splittedModel.length - 1];
+  // const suggestions = commands.filter(
```
davidcui-amzn, line 79: Are all these commented-out code blocks safe to remove?

public/components/common/seach/queryBar.tsx
```diff
@@ -9,28 +9,137 @@
...
 } from '../../../common/constants/explorer';
+import { Test } from 'mocha';
 
-export function QueryBar(props: IQueryBarProps) {
+const commands: EuiSuggestItemProps[] =[
+  {type: {iconType: 'search', color: 'tint1'} ,label: 'dedup'},
+  {type: {iconType: 'search', color: 'tint1'} ,label: 'eval'},
+  {type: {iconType: 'search', color: 'tint1'} ,label: 'fields'},
+  {type: {iconType: 'search', color: 'tint1'} ,label: 'rename'},
+  //{type: {iconType: 'search', color: 'tint1'} ,label: 'search'},
```
davidcui-amzn, line 37: Just curious, how come `search` as an option here is commented out and left in the `firstCommand` options while `source` is left in both?

sejli: I think `search` and `source` can both be first commands, but `source` be second if `search` is the first command. I was still working on the grammar portion, so I just left source in there for now. Will prob remove it

public/components/common/seach/queryBar.tsx
sejli: Yup, they were just leftover from previous commits.

davidcui-amzn: Got you- please remove these in your next commit then

public/components/common/seach/queryBar.tsx
```diff
@@ -72,53 +72,64 @@ export function QueryBar(props: IQueryBarProps) {
...
-      // return firstCommand.filter(
-      //   ( { label } ) => label.startsWith(prefix) && prefix !== label 
-      // );
     }
+    // TODO: (Grammar implementation) Get commands based on grammar from backend
+    // Contextual suggestions are hardcoded for now
     if (splittedModel.length > 1) {
-      // User just inputted a pipe
+      // Possible pipe commands
       if (splittedModel[splittedModel.length - 2] === "|"){
```
joshuali925, line 99: if user input is `source = index| head 10`, then `|` will not be on its own when you split by space?

sejli: forgot there could be spaces between `source` and `=`. i'm working on a more extensive grammar checker that is hardcoded for now and i'll remember to take this into account

public/components/common/seach/queryBar.tsx
```diff
@@ -9,42 +9,166 @@
...
+  const splittedModel = inputValue.split(' ');
+  const prefix = splittedModel[splittedModel.length - 1];
+
+  // Function to grab suggestions
+  const getSuggestions = () => {
+    // First commands should either be search or source
+    // source should be followed by an available index
+    if (splittedModel.length === 1) {
+      if (prefix.startsWith("source=")) {
+        let str = prefix.replace("source=", "");
```
davidcui-amzn, line 90: I don't see where `str` is modified after its declaration- can we declare it as a const?

sejli: yeah, just caught this recently. will change by my next commit

public/components/common/seach/queryBar.tsx
```diff
@@ -9,42 +9,166 @@
...
+        return getIndices(str);
+      }
+      return firstCommand;
+    }
+    // TODO: (Grammar implementation) Get commands based on grammar from backend
+    // Contextual suggestions are hardcoded for now
+    if (splittedModel.length > 1) {
```
davidcui-amzn, line 97: Can we make this second length check an `else if` block?

davidcui-amzn: Same with the `if` blocks within this conditional

sejli: will do

PR #81, opened 2021-07-30
Feature/autocomplete

sejli:
### Description
Switched to Algolia Autocomplete widget instead of EuiSuggest. Full functionality not quite there yet

### Issues Resolved

### Check List


public/components/common/seach/autocomplete.tsx
```diff
@@ -0,0 +1,278 @@
+/*
```
davidcui1225, line 1: Can we remove this first SPDX License Identifier? looks like a dup

public/components/common/seach/autosuggest.tsx
```diff
@@ -0,0 +1,28 @@
+import { autocomplete } from '@algolia/autocomplete-js';
```
davidcui1225, line 1: Missing license identifier

public/components/common/seach/autocomplete_service.tsx
```diff
@@ -0,0 +1,83 @@
...
+        {label: 'where'},
+        {label: 'head'},
+        {label: 'where'},
+        {label: 'top'}
+    ]
+const getQuerySuggestions = () => {
+    
+    const [searchValue, setSearchValue] = useState('');
+    const [isLoading, setLoading] = useState(false);
+    //const [searchRef, setSearchRef] useState<HTMLInputElement | null>(null);
```
davidcui1225, line 34: Is this commented-out code safe to remove?

sejli: yeah, i think this was eric's old changes. i'll get remove it

public/components/common/seach/autosuggest.tsx
sejli: think this was a dupe file, will delete

public/components/common/seach/autocomplete.tsx
```diff
@@ -0,0 +1,278 @@
...
+
+const statsCommands = [
+  { label: 'count()' },
+  { label: 'sum()' },
+  { label: 'avg()' },
+  { label: 'max()' },
+  { label: 'min()' },
+];
+
+const indices = [
```
davidcui1225, line 53: How come we are hard-coding the list of indices?

public/components/common/seach/autocomplete_service.tsx
mengweieric: This wasn't my code change at all. This is a new file you crated that doesn't exist in any other branches.

public/components/common/seach/autocomplete.tsx
davidcui1225: I guess it seems like everything is hard-coded. Follow-up question, where will we pull the options from for each of the current hard-coded consts once we have full functionality?

public/components/common/seach/autocomplete.tsx
```diff
@@ -0,0 +1,270 @@
...
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+import './search.scss';
+import React, { createElement, Fragment, useEffect, useRef, useState } from 'react';
+import { render } from 'react-dom';
+import { h } from 'preact';
```
joshuali925, line 14: what's preact

public/components/common/seach/autocomplete.tsx
```diff
@@ -0,0 +1,270 @@
...
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+import './search.scss';
+import React, { createElement, Fragment, useEffect, useRef, useState } from 'react';
+import { render } from 'react-dom';
```
joshuali925, line 13: i don't think this is used?

public/components/common/seach/autocomplete.tsx
```diff
@@ -0,0 +1,270 @@
...
+
+let inFieldsCommaLoop = false;
+let hasFields = false;
+const fieldsFromBackend = [];
+
+// Function to grab suggestions
+export function getSuggestions(str: string, http) {
+  const splittedModel = str.split(' ');
+  const prefix = splittedModel[splittedModel.length - 1];
+  let itemSuggestions = [];
```
joshuali925, line 59: maybe no need to initialize to `[]`?

public/components/common/seach/autocomplete.tsx
```diff
@@ -0,0 +1,270 @@
...
+  { label: 'opensearch_dashboards_sample_data_logs' },
+];
+
+let inFieldsCommaLoop = false;
+let hasFields = false;
+const fieldsFromBackend = [];
+
+// Function to grab suggestions
+export function getSuggestions(str: string, http) {
+  const splittedModel = str.split(' ');
```
joshuali925, line 57: just a question on the high level logic, PPL is separated by pipes, why split on space instead of `|`? splitting by | should give you a list of commands which can then be parsed on words or chars

public/components/common/seach/autocomplete_service.tsx
```diff
@@ -0,0 +1,70 @@
...
+    EuiFlexItem,
+    EuiListGroup,
+    EuiListGroupItem,
+    EuiText,
+    EuiSelectableTemplateSitewide,
+    EuiSelectableTemplateSitewideOption,
+    EuiLink,
+    EuiFlexGroup
+  } from '@elastic/eui';
+  import {
```
joshuali925, line 13: not a problem but some indentation seems off, could you run a formatter over these files? i think we are mostly using prettier as formatter

public/components/common/seach/autocomplete.tsx
mengweieric: Sean is currently working on getting fields and respective values for autocomplete, those works will be added later in the future commits. Previously he was focusing on the UI works, and how to get data for suggestions was not yet determined at that time. Therefore in order to demo the progress, test data was hardcoded for UI functionalities. 

Matched values for a field will be fetched through DSL (Sean is working on). Commands, functions and other PPL related symbols will be fetched through the ppl plugin (opensearch side to-do item).

As for fetching index patterns, it depends on a saved-object solution in order to save/ get all indices for suggestions, which is also a to-do item.

public/components/common/seach/autocomplete.tsx
mengweieric: Do we need preact? I thought dangerouslySetInnerHTML is good enough to add html.

public/components/common/seach/autocomplete_service.tsx
sejli: Oh right my bad. I completely forgot i wrote this (think it was when i was trying out different widgets). I'll remove this file altogether because it's redundant now

public/components/common/seach/autocomplete.tsx
sejli: Nope, we don't need it. Will remove

public/components/common/seach/autocomplete.tsx
sejli: Not used, will remove

public/components/common/seach/autocomplete_service.tsx
sejli: Will be getting rid of these files, but I've run eslint over the other ones

public/components/common/seach/autocomplete.tsx
sejli: I split on space to list suggestions for each specific word. Since there's only a few commands, I have them all in an array and whenever the user types out a word, the array of commands is filtered by the most recent word. I also wanted to be able to check previous words; for example, `search` must be followed by `source =`, etc. I've just been adding more checks to my giant if else block as I continue with the project, but the plan is to refactor the grammar check to something more robust and coherent (split by pipes, then for the first element in the array the user must type search or source and an index? use properties to keep track of selected index or commands? something like that). Definitely open to suggestions, I've been adding checks to my code as I go along.

public/components/common/search/autocomplete.tsx
```diff
@@ -0,0 +1,270 @@
...
+import './search.scss';
+import React, { createElement, Fragment, useEffect, useRef, useState } from 'react';
+
+import { autocomplete } from '@algolia/autocomplete-js';
+import { RAW_QUERY } from '../../../common/constants/explorer';
+import { IQueryBarProps } from './search';
+import { handlePplRequest } from '../../../requests/ppl';
+
+// Possible suggestions (hardcoded)
+const firstCommand = [{ label: 'search' }, { label: 'source' }];
```
joshuali925, line 20: i think i've mostly seen PPL starting with `source`, is there a use case to have it starts with `search`?

public/components/common/search/autocomplete.tsx
```diff
@@ -0,0 +1,270 @@
...
+// Function to grab suggestions
+export function getSuggestions(str: string, http) {
+  const splittedModel = str.split(' ');
+  const prefix = splittedModel[splittedModel.length - 1];
+  let itemSuggestions = [];
+  const fullSuggestions = [];
+
+  // First commands should either be search or source
+  // source should be followed by an available index
+  if (prefix.startsWith('source=')) {
```
joshuali925, line 62: would this work if user inputs spaces? like
```
source =  index
```

public/components/common/search/autocomplete.tsx
```diff
@@ -0,0 +1,270 @@
...
+      fullSuggestions.push({
+        label: str.substring(0, str.lastIndexOf(prefix)) + itemSuggestions[i].label,
+        input: str,
+        suggestion: itemSuggestions[i].label.substring(prefix.length),
+      });
+    }
+    return fullSuggestions;
+  }
+
+  if (prefix.includes(';')) {
```
joshuali925, line 79: what if semicolon is part of the query? like `source = accounts employer != 'non-standard;name'`

public/common/types/explorer.ts
```diff
@@ -55,6 +55,7 @@ export interface IExplorerFields {
 export interface IExplorerProps {
   tabId: string;
   query: any;
+  http: any;
```
joshuali925, line 58: i think `http` is `HttpStart` in here `https://github.com/opensearch-project/OpenSearch-Dashboards/blob/1dc90f5d16316fcc49a7ef67a9e6306a407cee59/src/core/public/http/types.ts#L90:13`, might be better to reduce the number of `any` types

public/components/common/seach/autocomplete.tsx
joshuali925: I see. even splitting by `|` has a possibility to break if user has something like `source=index field="a|b"`. I think many parsers/compilers reads by character into tokens, but i'm not sure if full parsing of the query is needed for autocomplete and i have zero experience in these before. just worried that the big `if-else` implementation could get too complicated to maintain, just a thought

PR #83, opened 2021-08-06
Feature/visualization

mengweieric:
### Description
Code changes for redux, and parts of the visualization 

### Issues Resolved
Redux store


package.json
```diff
@@ -10,7 +10,11 @@
     "node": "10.23.1",
```
joshuali925, line 10: could you remove this block, also add some scripts (mainly `osd` and `build`), reference here https://github.com/opensearch-project/notifications/blob/develop/dashboards-notifications/package.json#L12-L20

public/components/common/debounced_component/debounced_component.test.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+  test('debounces changes', async () => {
+    const TestComponent = debouncedComponent(({ title }: { title: string }) => {
+      return <h1>{title}</h1>;
+    }, 1);
+    const component = mount(<TestComponent title="there" />);
+    component.setProps({ title: 'yall' });
+    expect(component.text()).toEqual('there');
+    await act(async () => {
+      await new Promise((r) => setTimeout(r, 1));
```
joshuali925, line 33: not very familiar with the event loop, but would it be possible that this finishes before debounced component re-renders and test fails? since the delays are both `1`

public/components/explorer/visualizations/drag_drop/drag_drop.tsx
```diff
@@ -0,0 +1,255 @@
...
+type Props = DraggableProps | NonDraggableProps;
+
+/**
+ * A draggable / droppable item. Items can be both draggable and droppable at
+ * the same time.
+ *
+ * @param props
+ */
+
+export const DragDrop = (props: Props) => {
```
joshuali925, line 121: i think there are existing react libraries that can handle drag drop, any reason to implement it instead?

public/components/explorer/visualizations/datapanel.tsx
```diff
@@ -0,0 +1,74 @@
...
+      className="lnsInnerIndexPatternDataPanel"
+      direction="column"
+      responsive={false}
+    >
+      <EuiFlexItem grow={false}>
+        <EuiFormControlLayout
+          icon="search"
+          fullWidth
+          clear={{
+            title: i18n.translate('xpack.lens.indexPatterns.clearFiltersLabel', {
```
joshuali925, line 41: i don't think xpack should be here?

public/common/types/explorer.ts
```diff
@@ -49,18 +49,12 @@ export interface IExplorerTabFields {
...
 export interface IExplorerProps {
   tabId: string;
-  query: any;
-  explorerData: any;
-  explorerFields: any;
-  setSearchQuery: (query: string, tabId: string) => void;
-  querySearch: (tabId: string) => void;
-  addField: (field: IField, tabId: string) => void;
-  removeField: (field: IField, tabId: string) => void
+  pplService: any;
```
joshuali925, line 57: the type should just be `PPLService`?

joshuali925 (approved):
minor comments

package.json
mengweieric: Sure, will do.

public/common/types/explorer.ts
mengweieric: Yes, missed this one.

public/components/explorer/visualizations/drag_drop/drag_drop.tsx
mengweieric: I was borrowing code directly from kibana lens, since the design of visualization in explorer looks similar to it and sharing some of the functionalities. But moving forward, I'll take look at react drag and drop libraries to see what could be the best solution for this feature of visualization.

public/components/explorer/visualizations/datapanel.tsx
mengweieric: Same thing here as I was borrowing code from existing kibana lens and only made minimal changes. I'll remove these unrelated stuff.

public/components/common/debounced_component/debounced_component.test.tsx
mengweieric: To me it looks like it would not fail since when setting the props, this action would be put into the event loop, then the following assertion statement would run immediately and pass. No matter what statement is actually picked by the main thread to run next (setting props to 'yall' or await act...), it would actually end up making sure the debounce component has changed its value to 'yall', before the last assertion runs. Since event handlers will be invoked in the same order that they were bound in.

joshuali925: got it, looks good then

public/components/common/debounced_component/index.ts
```diff
@@ -0,0 +1,12 @@
...
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+export * from './debounced_component';
```
davidcui1225, line 12: Can we avoid the wild card import usage here?

public/components/explorer/hooks/index.ts
```diff
@@ -0,0 +1,12 @@
...
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+export * from './useFetchQueryResponse';
```
davidcui1225, line 12: Same question here

public/components/explorer/timechart_header/timechart_header.tsx
```diff
@@ -87,9 +88,9 @@ export function TimechartHeader({
     onChangeInterval(e.target.value);
   };
 
-  if (!timeRange || !bucketInterval) {
-    return null;
-  }
+  // if (!timeRange || !bucketInterval) {
```
davidcui1225, line 91: Can we delete these commented-out code blocks?

davidcui1225 (approved):
LGTM- left a few minor comments

public/components/explorer/timechart_header/timechart_header.tsx
mengweieric: Sure, will do.

public/components/common/debounced_component/index.ts
mengweieric: Ok, will change all of them

Issue #78, opened 2021-07-29
Observability - supports rendering visualization based on PPL query


mengweieric:
**Is your feature request related to a problem? Please describe.**
Previously it is using only the test data to render bar/line chart.

**Describe the solution you'd like**
PPL now supports aggregation by specifying timespan and returning data format that is favored by plotly.js, the UI should add code to use new data format.

Issue #53, opened 2021-06-16
Observability : Event Analytics - Visualizations


mengweieric:
**Is your feature request related to a problem? Please describe.**
Add Charting support to PPL and event analytics

<img width="916" alt="Screen Shot 2021-08-09 at 9 50 03 AM" src="https://user-images.githubusercontent.com/19362/128743524-96a5f7b4-27d5-4fd1-aedb-f852124a61e4.png">
<img width="765" alt="Screen Shot 2021-08-09 at 9 50 46 AM" src="https://user-images.githubusercontent.com/19362/128743634-3dd6492d-4b1c-4e45-8600-f18a5a95ae69.png">

**Describe the solution you'd like**
Top 3 most popular charts in chart tab.

anirudha:
https://user-images.githubusercontent.com/19362/137158907-bf5814ae-59d3-4f61-8ecf-649f49c17560.mov

PR #88, opened 2021-08-18
Feature/visualization

mengweieric:
### Description
Added visualization related features.

### Issues Resolved
- Supports rendering visualization based on PPL query [#78]
- Explorer visualization panel [#53]

### Additional Notes
- Having intermittent issues with legacy opensearch client when processing concurrent requests, will fix it later.
- Query/visualization history is dummy now for demo purpose


joshuali925 (approved):

public/common/types/explorer.ts
```diff
@@ -22,9 +22,9 @@ export interface IQueryTab {
   content: React.ReactNode
 }
 
-export interface ILogExplorerProps {
-  http: CoreStart['http']
-}
+// export interface ILogExplorerProps {
```
davidcui1225, line 25: Safe to remove the commented export here?

mengweieric: yes, I'll remove.

public/components/explorer/explorer.tsx
```diff
@@ -94,6 +113,20 @@ export const Explorer = ({
     [setFixedScrollEl]
   );
 
+  const getIndexFromRawQuery = (rawQuery: string) => rawQuery.split('=')[1].split(' ')[0];
+
+  useEffect(() => {
+    if (!query) return;
+    if (query.includes('stats')) {
```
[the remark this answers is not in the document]
davidcui1225: +1, had a similar check in dashboards-reports that ended up causing an issue and we had to change it to a stricter check.

public/components/explorer/visualizations/shared_components/toolbar_popover.tsx
```diff
@@ -50,7 +50,7 @@ export const ToolbarPopover: React.FunctionComponent<ToolbarPopoverProps> = ({
   const [open, setOpen] = useState(false);
 
   const iconType: string | IconType = typeof type === 'string' ? typeToIconMap[type] : type;
-
+  console.log('iconType: ', iconType);
```
davidcui1225, line 53: np: remove console.log statement

server/adaptors/pplDatasource.ts
```diff
@@ -43,21 +42,11 @@ export class PPLDataSource {
...
-          ) {
-            hasTimestamp = true;
-        }
       }
 
       data.push(record);
     });
-    pplRes['jsonData'] = data;
-    pplRes['hasTimestamp'] = hasTimestamp;
+    this.pplDataSource['jsonData'] = data;
```
davidcui1225, line 49: What are the parameters of the `pplDataSource` object, how come `jsonData` is being called with bracket notation vs. dot notation? Should we add a Type or Interface for it?

public/components/explorer/explorer.tsx
```diff
@@ -94,6 +113,20 @@ export const Explorer = ({
     [setFixedScrollEl]
   );
 
+  const getIndexFromRawQuery = (rawQuery: string) => rawQuery.split('=')[1].split(' ')[0];
```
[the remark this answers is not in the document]
mengweieric: Yes, this makes sense. We have plan to make autocomplete add index pattern to state data whenever it is selected through autosuggestion, and access the index directly without processing raw query string. I think once that part is finished, then here we don't need to split the string in order to get the index. But still we'll need to cover corner cases where user just paste a query without selecting index, then it needs to parse the raw query. I'll use regex.

public/components/explorer/explorer.tsx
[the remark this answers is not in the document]
mengweieric: Yes, I'll use regex.

public/components/explorer/visualizations/shared_components/toolbar_popover.tsx
mengweieric: Right, thanks, I'll remove.

server/adaptors/pplDatasource.ts
mengweieric: The pplDataSource is just the response object returned from opensearch with schema-to-value mapping added. I'll add interface for it.

PR #94, opened 2021-08-24
Feature query regex

### LINKED ISSUE #91: Using Regex to check 'stats' keyword in and extract index patterns from raw query.
**Is your feature request related to a problem? Please describe.**
Currently, it uses includes() to check 'stats' command and split() to get index patterns which is not a stable and bug-free solution

**Describe the solution you'd like**
Use regular expression to match patterns.

mengweieric:
### Description

- Use regex to retrieve index from query
- File renaming to unify naming conventions
- Restructured project

### Issues Resolved
#91 


common/utils/query_utils.ts
```diff
@@ -0,0 +1,18 @@
...
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+export const getIndexPatternFromRawQuery = (query: string) => {
+  const indexMatches = query.match('\=(.*)');
+  if (indexMatches) {
+    return indexMatches[0]?.slice(1).split('|')[0]?.trim();
```
ps48: Suggestion: is there a regex way to do this too?

ps48 (approved):
Minor changes. LGTM

common/utils/query_utils.ts
```diff
@@ -0,0 +1,18 @@
...
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ *
+ * Modifications Copyright OpenSearch Contributors. See
+ * GitHub history for details.
+ */
+
+export const getIndexPatternFromRawQuery = (query: string) : string => {
+  const indexMatches = query.match('\=(.*)');
```
joshuali925, line 13: maybe use `/regex/` instead of `'regex'`, also the `\` and `()` doesn't seem to make a difference? 

does this work with the example i gave before? I was thinking something like `.match(/source\s*=\s*([^|\s]+)/)[1]` instead of js string manipulation

public/components/explorer/explorer.tsx
```diff
@@ -113,18 +116,21 @@ export const Explorer = ({
...
   );
 
-  const getIndexFromRawQuery = (rawQuery: string) => rawQuery.split('=')[1].split(' ')[0];
-
-  useEffect(() => {
+  const fetchData = () => {
     if (!query) return;
-    if (query.includes('stats')) {
-      const index = getIndexFromRawQuery(query);
+    if (statsRegx.test(query)) {
```
joshuali925, line 121: similar here, this is the same as `query.includes('stats')`. maybe check if `stats` is followed directly (or with some spaces) after `|`, or some other conditions that's only true for the `stats` command?

public/components/explorer/explorer.tsx
```diff
@@ -113,17 +116,24 @@ export const Explorer = ({
...
   );
 
-  const getIndexFromRawQuery = (rawQuery: string) => rawQuery.split('=')[1].split(' ')[0];
+  // const getIndexFromRawQuery = (rawQuery: string) => {
+  //   const indexMatches = rawQuery.match('\=(.*)');
+  //   if (indexMatches) {
+  //     return indexMatches[0]?.slice(1).split('|')[0]?.trim();
+  //   }
+  //   return '';
+  // }
```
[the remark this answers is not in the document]
mengweieric: These comments have been removed in later commits 426b118.

common/utils/query_utils.ts
mengweieric: Sure, we can use /regex/. Yes, the \ and () can be removed in this case. 

This works with the example you gave before and also other corner cases. But I agree it's better to use regex entirely in this case.

public/components/explorer/explorer.tsx
mengweieric: Sure, I'm thinking to use query.match(/\|\s*stats/) here to match 'stats'.

common/utils/query_utils.ts
joshuali925: Got it, then looks good. up to you on which implementation you want to use.

just a note i did not test the regex i wrote. if you use something similar maybe also add the `/i` flag to ignore case, and i think `index` can be used instead of `source` so maybe `(source|index)`

public/components/explorer/explorer.tsx
mengweieric: yea, makes sense, I'm thinking and testing query.match(/\|\s*stats/i) to match 'stats'

common/utils/query_utils.ts
mengweieric: Exactly, I'm adding index and /i to it, thanks

public/components/explorer/explorer.tsx
joshuali925: cool, i think the `|` character might need escape

mengweieric: yes, the one I used includes the escape, I noticed that when I left the comment it was missing that escape. I tried to modified it to add escape but it just not reflecting it. Wired.

common/utils/query_utils.ts
```diff
@@ -9,10 +9,4 @@
...
 
-export const getIndexPatternFromRawQuery = (query: string) : string => {
-  const indexMatches = query.match('\=(.*)');
-  if (indexMatches) {
-    return indexMatches[0]?.slice(1)?.split('|')[0]?.trim();
-  }
-  return '';
-};
\ No newline at end of file
+export const getIndexPatternFromRawQuery = (query: string) : string => query.match(/(source|index)\s*=\s*([^|\s]+)/i)[1];
```
joshuali925, line 12: i think the captured group index should be 2? since you added another one in front. Also maybe add some guard if no match? `query.match(...)?.[2] ?? ''` (or use a variable and add null check, i don't know does `?.` make regex match run one more time?)

mengweieric: Yea it should be [2] but interesting though is that the app is running correctly with match...tried to reboot but still...and after the change it is also working. Anyway I changed it to use the [2] for matching, thanks for catching this. Yea, I think if there's no match the returned res would be null.

joshuali925: hmm interesting, [1] should not work, and query.match() returns null if there is no match and indexing null will crash the application. your browser cache might be outdated, try "empty cache and hard reload" or incognito on browser?

mengweieric: yea cache issue probably, now it's broken. but I used 'empty cache and hard reload' all the time...hmmm

joshuali925 (approved):

Issue #100, opened 2021-08-31
Custom Panels - Visualization Routes


ps48:
**Is your feature request related to a problem? Please describe.**
Support server routes for visualization

**Describe the solution you'd like**
Add APIs for the following:
1. Add a visualization
2. Replace a visualization
3. Clone a visualization
4. Remove a visualization
5. Edit visualization size and position

Issue #99, opened 2021-08-31
Custom Panels - Panel Routes 


ps48:
**Is your feature request related to a problem? Please describe.**
Support for panel based server routes 

**Describe the solution you'd like**
Support panel APIs for:
1. Fetch all panels
2. Fetch panel by id
3. Create a new panel
4. Rename a panel
5. Clone a panel
6. Delete a panel
7. Export a panel
8. Apply date and query filter to a panel

Issue #90, opened 2021-08-23
Custom Panels - Add PPL filter Support 


ps48:
**Is your feature request related to a problem? Please describe.**
Allow users to add PPL queries in the header bar to apply filter on all visualizations present in the panel.

**Describe the solution you'd like**
Integrate the header PPL query with the existing data query present in a visualization.

Issue #75, opened 2021-07-27
Custom Panels - Visualization Actions Menu 


ps48:
**Is your feature request related to a problem? Please describe.**
Custom Observability Panels - Visualization Actions Menu 

**Describe the solution you'd like**
1. Allow users to Edit/Duplicate/Remove preexisting visualizations on panel
2. Users should see a model to confirm the removal of visualization
3. Allow users to change the width of visualization (33%, 50%  or 100% of the window)

PR #108, opened 2021-09-08
Added ppl query filter, added router placeholder for panels

### LINKED ISSUE #77: Custom Panels - Query and Filtering
**Is your feature request related to a problem? Please describe.**
Custom Observability Panels - Query and Filtering 

**Describe the solution you'd like**
1. Enable users to query and filter all visualizations using PPL commands
2. Allow users to change date and time range of all the visualizations in the panel

ps48:
### Description
Features added:
1. Panel Query Filter
2. Visualization Actions menu (Remove, Clone)
2. Router placeholder for panel APIs
3. Router placeholder for visualization APIs

### Issues Resolved
#77 
#90 
added placeholder for #99 & #100 
Partial #75 


common/constants/shared.ts
```diff
@@ -24,3 +24,10 @@ export const observabilityPluginOrder = 6000;
 export const UI_DATE_FORMAT = 'MM/DD/YYYY hh:mm A';
 export const PPL_DATE_FORMAT = 'YYYY-MM-DD HH:mm:ss';
 export const PPL_INDEX_REGEX = /(search source|source|index)\s*=\s*([^|\s]+)/i;
+
+// Observability plugin URI 
+const BASE_OBSERVABILITY_URI = '/_plugins/_observability';
```
joshuali925, line 29: are we refactoring notebooks backend plugin to provide this? or is it a new plugin

public/components/custom_panels/panel_modules/panel_grid.tsx
```diff
@@ -50,9 +53,12 @@ export const PanelGrid = ({
   startTime,
   endTime,
   onRefresh,
+  cloneVisualization,
+  deleteVisualization,
+  pplFilterValue,
 }: Props) => {
-  const [layout, setLayout] = useState([]);
-  const [editedLayout, setEditedLayout] = useState([]);
+  const [layout, setLayout] = useState([] as Layout[]);
```
joshuali925, line 60: not an issue but i think the syntax is usually `useState<Layout[]>([])`?

server/adaptors/custom_panels/custom_panel_adaptor.ts
```diff
@@ -0,0 +1,93 @@
...
+      const response = await client.callAsCurrentUser('observability.getPanelById', {
+        panelId: panelId,
+      });
+      return response.panelDetails;
+    } catch (error) {
+      throw new Error('Get Panel Error:' + error);
+    }
+  };
+
+  // gets first `FETCH_SIZE` of panels available
```
joshuali925, line 65: is this comment still applicable?

joshuali925 (approved):

common/constants/shared.ts
ps48: I'm not sure about this. One thing would be, backward compatibility to keep in mind. @anirudha  had mentioned about keeping all documents in observability indexed together(differentiated by types) something similar to how ".kibana" index does.

public/components/custom_panels/panel_modules/panel_grid.tsx
ps48: will change this

server/adaptors/custom_panels/custom_panel_adaptor.ts
ps48: will change this as well :D

server/plugin.ts
```diff
@@ -42,7 +43,7 @@ export class ObservabilityPlugin implements Plugin<ObservabilityPluginSetup, Obs
     core.http.registerRouteHandlerContext('observability_plugin', (context, request) => {
       return {
         logger: this.logger,
-        observabilityClient,
+        observabilityClient: OpenSearchObservabilityPlugin,
```
joshuali925, line 46: This line replaces `observabilityClient` object with `OpenSearchObservabilityPlugin` in plugin context, I think what you want is to add `OpenSearchObservabilityPlugin` after `PPLPlugin` in line 39?

I don't think `observabilityClient` is being accessed through context anywhere so this doesn't break anything yet, but just want to double check if this is intentional @ps48

ps48: Oh yes, that was intentional. The backend is not fully done so was waiting to finish the Opensearch Part and then add it here. But if for notebooks integration you want to add this, please go ahead.

joshuali925: But why expose `OpenSearchObservabilityPlugin` in router context? it's a helper function to construct a client. caller needs to provide `Client` so it defines client actions in the Client prototype, i don't think you can use it directly?

I might need to modify this for notebooks to work, since current setup needs `context.observability_plugin.observabilityClient` to be a client, and I think we should only have one `observabilityClient`

ps48: I think, I was about to change the name of Client object name [observabilityClient => openSearchObservabilityClient], but somehow forgot to do that. And yes, line 46 should be  `observabilityClient: openSearchObservabilityClient,`. For notebooks you can add your API endpoints in the common `OpenSearchObservabilityPlugin`

PR #130, opened 2021-10-12
Feature/operational panels backend

ps48:
### Description
Added Backend APIs to the Operational Panels using Observability plugin

### Issues Resolved
1. Added Operational Panels Kibana APIs #99 #100 
2. Event Explorer API endpoint for adding visualizations to any panel

### DIFF
--- a/server/adaptors/custom_panels/custom_panel_adaptor.ts
+++ b/server/adaptors/custom_panels/custom_panel_adaptor.ts
@@ -9,9 +9,12 @@
  * GitHub history for details.
  */
 
-import { PanelType, VisualizationType } from '../../../common/types/custom_panels';
-import { ILegacyScopedClusterClient } from '../../../../../src/core/server';
-import { PPL_CONTAINS_TIMESTAMP_REGEX } from '../../../common/constants/shared';
+import {
+  PanelType,
+  VisualizationType,
+} from "../../../common/types/custom_panels";
+import { ILegacyScopedClusterClient } from "../../../../../src/core/server";
+import { PPL_CONTAINS_TIMESTAMP_REGEX } from "../../../common/constants/shared";
 
 export class CustomPanelsAdaptor {
   // index a panel
@@ -20,14 +23,17 @@ export class CustomPanelsAdaptor {
     panelBody: PanelType
   ): Promise<{ objectId: string }> {
     try {
-      const response = await client.callAsCurrentUser('observability.createObject', {
-        body: {
-          operationalPanel: panelBody,
-        },
-      });
+      const response = await client.callAsCurrentUser(
+        "observability.createObject",
+        {
+          body: {
+            operationalPanel: panelBody,
+          },
+        }
+      );
       return response;
     } catch (error) {
-      throw new Error('Index Panel Error:' + error);
+      throw new Error("Index Panel Error:" + error);
     }
   };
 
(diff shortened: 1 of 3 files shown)


common/constants/shared.ts
```diff
@@ -30,6 +30,7 @@ export const observabilityPluginOrder = 6000;
 export const UI_DATE_FORMAT = 'MM/DD/YYYY hh:mm A';
 export const PPL_DATE_FORMAT = 'YYYY-MM-DD HH:mm:ss';
 export const PPL_INDEX_REGEX = /(search source|source|index)\s*=\s*([^|\s]+)/i;
+export const PPL_CONTAINS_TIMESTAMP_REGEX = /\|\s*.*\s*[<|<=|=|>=|>]\s*timestamp\([^\)]+\)/i;
```
mengweieric, line 33: Do you need this timestamp regex in custom panel? This was added previously but the new requirement asks to not check if a query contains a date range anymore, I've already removed on my side.

mengweieric (approved):

server/adaptors/custom_panels/custom_panel_adaptor.ts
```diff
@@ -50,44 +50,335 @@ export class CustomPanelsAdaptor {
...
+        objectType: 'operationalPanel',
+      });
+      return response.observabilityObjectList.map((panel: any) => ({
+        name: panel.operationalPanel.name,
+        id: panel.objectId,
+        dateCreated: panel.createdTimeMs,
+        dateModified: panel.lastUpdatedTimeMs,
       }));
     } catch (error) {
       if (error.body.error.type === 'index_not_found_exception') {
```
joshuali925, line 78: you can remove this condition, the backend will handle it

server/adaptors/custom_panels/custom_panel_adaptor.ts
```diff
@@ -50,44 +50,335 @@ export class CustomPanelsAdaptor {
...
+      });
+      return response.observabilityObjectList.map((visualization: any) => ({
+        id: visualization.objectId,
+        name: visualization.savedVisualization.name,
+        query: this.savedVisualizationsQueryBuilder(visualization.savedVisualization.query),
+        type: visualization.savedVisualization.type,
+        timeField: visualization.savedVisualization.selected_timestamp.name,
+      }));
+    } catch (error) {
+      if (error.body.error.type === 'index_not_found_exception') {
```
joshuali925, line 224: same here

server/routes/custom_panels/panels_router.ts
```diff
@@ -415,16 +275,26 @@ export function PanelsRouter(router: IRouter) {
...
+        );
         return response.ok({
           body: {
-            message: 'Panel DateTime Filter Changed',
+            message: 'Panel PPL Filter Changed',
           },
         });
       } catch (error) {
-        console.log('Issue in adding datetime filter', error);
+        console.log('Issue in adding query filter', error);
```
joshuali925, line 297: maybe use `console.error`?

common/constants/shared.ts
ps48: This is added here cause, I still need to remove time range from queries in "savedVisualization" data modal. 
1. If we are still allowing the users to add time range in query in event analytics, then I need to remove them explicitly in panels' visualizations.
2. If we are not allowing users to add time range in query in event analytics, then I'll remove this check. 
Please let me know if got something wrong here.

ps48: Oh yes, I understand this point that date range is stored separately and only added in runtime. But, I would like to know what happens if the user adds a time range in the query? Do we explicitly remove? Or Do we throw an error toast saying "adding time range is not allowed in query"? If we have such checks I'll remove the checks from my side.

PR #190, opened 2021-11-04
Fix for duplicate indices in suggestion

eugenesk24:

### Description
Duplicate items are filtered out before they render. Background color of suggestion dropdown changes to dark mode. 

### Issues Resolved
Suggestions for indices were showing up as double or triple. 

### Check List


dashboards-observability/public/components/common/search/autocomplete.tsx
```diff
@@ -359,6 +359,7 @@ export function Autocomplete({
         {autocompleteState.isOpen &&
           autocompleteState.collections.map((collection, index) => {
             const { source, items } = collection;
+            items.filter((item, index) => { return items.indexOf(item) === index })
```
ps48, line 362: Do we know the root cause of duplicates? 
* Is it like making two requests?
* Browser caching?
* Or the variable is not being cleared?

ps48 (approved):

dashboards-observability/public/components/common/search/autocomplete.tsx
eugenesk24: Not 100% sure but there is only one request being made and the variable should only be updated once because it's in an if condition that it must be empty for it to be updated. Don't know how to check the browser caching.

eugenesk24: Wait, just kidding. I just checked and it looks like there are two requests being made.

ps48 (approved):

PR #207, opened 2021-11-09
Case insensitive Autocomplete

eugenesk24:

### Description
Autocomplete is now case insensitive when giving suggestions

### Issues Resolved
Users had to type the correct casing in order to see matching suggestions.

### Check List

### DIFF
--- a/dashboards-observability/public/components/common/search/autocomplete.tsx
+++ b/dashboards-observability/public/components/common/search/autocomplete.tsx
@@ -94,6 +94,7 @@ const getFirstPipe = async (str: string, dslService: DSLService) => {
 const getSuggestions = async (str: string, dslService: DSLService) => {
   const splittedModel = str.split(' ');
   const prefix = splittedModel[splittedModel.length - 1];
+  const lowerPrefix = prefix.toLowerCase();
   const fullSuggestions: any = [];
 
   // Check the last full word in the query, then suggest inputs based off that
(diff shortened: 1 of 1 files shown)


dashboards-observability/public/components/common/search/autocomplete.tsx
```diff
@@ -67,7 +67,7 @@ const statsCommands = [
 // Function to create the array of objects to be suggested
 const fillSuggestions = (str: string, word: string, items: any) => {
   const filteredList = items.filter(
-    (item: { label: string }) => item.label.startsWith(word) && word !== item.label
+    (item: { label: string }) => item.label.toLowerCase().startsWith(word.toLowerCase()) && word.toLowerCase().localeCompare(item.label.toLowerCase())
```
joshuali925, line 70: why is both `startsWith` and `localeCompare` needed? Also maybe assign a variable to `word.toLowerCase()` so it's not calculated every loop

eugenesk24: There were some instances where !== was returning true for words that were the same, this seemed to fix that issue. An example was "Dest" and "dest". I'll make the variable though!

joshuali925: i meant if you are comparing for equality, then `startsWith` is always true, is it necessary to check both?

eugenesk24: We check for both because we want to keep it if the label starts with the word but not keep it if they are the same. 
So if the user types "Dest" we want to suggest "DestCityName", "DestCountry" but not "Dest"

joshuali925: oh you are using `!!localeCompare(..) === false` to check if it's equal, i feel it's a bit confusing but ok

joshuali925 (approved):

dashboards-observability/public/components/common/search/autocomplete.tsx
eugenesk24: Oh yea, localeCompare returns 0 if they are equal which is definitely confusing. Should I add a comment there?

PR #208, opened 2021-11-09
Feature autocomplete fix

mengweieric:
### Description
- autocomplete highlighting issue fix
- home page enhancements

### Issues Resolved

### Check List
- [ x ] Commits are signed per the DCO using --signoff

### DIFF
--- a/dashboards-observability/public/components/common/search/autocomplete.tsx
+++ b/dashboards-observability/public/components/common/search/autocomplete.tsx
@@ -23,10 +23,9 @@ import {
 import { EuiTextArea } from '@elastic/eui';
 import { IQueryBarProps } from './search';
 import { getDataValueQuery } from './queries/data_queries';
-import { isEmpty, isEqual } from 'lodash';
+import { isEmpty } from 'lodash';
 import DSLService from 'public/services/requests/dsl';
 import { uiSettingsService } from '../../../../common/utils';
-import { RAW_QUERY } from 'common/constants/explorer';
 
 let currIndex: string = '';
 let currField: string = '';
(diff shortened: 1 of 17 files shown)


dashboards-observability/public/components/common/search/autocomplete.tsx
```diff
@@ -178,11 +180,14 @@ const getSuggestions = async (str: string, dslService: DSLService) => {
...
-        await getDataValues(currIndex, currField, currFieldType, dslService)
-      );
+      if (isEmpty(prefix)) {
+        return fillSuggestions(
+          str,
+          prefix,
+          await getDataValues(currIndex, currField, currFieldType, dslService)
+        );
+      }
+      return [];
```
eugenesk24, line 190: Wouldn't this wipe the suggestions after the user types the first character of the datavalue?

mengweieric: Previously it simply just kept sending requests to retrieve values for a field on every typing. The existing implementation would cause errors/issues since it needs to change to use keyword rather than text for search. I checked with Ani on this yesterday, he's asking to disable the call when user typed in at least one character. I think this is going to be just a quick fix to prevent from overwhelming the backend and having many errors on each typing. But I think we would change the implementation later.

mengweieric: Previously it simply just kept sending requests to retrieve values for a field on every typing. The existing implementation would cause errors/issues since it needs to change to use keyword rather than text for search. I checked with Ani on this yesterday, he's asking to disable the call when user typed in at least one character. This is going to be just a quick fix to prevent from overwhelming the backend and having many errors on each typing. But I think we would change the implementation later.

ps48 (approved):

anirudha (approved):

joshuali925 (approved):

PR #265, opened 2021-11-18
fixed tab close issue

mengweieric:
Signed-off-by: Eric Wei <menwe@amazon.com>

### Description
- Tab close issue
- Resolved a warning on EuiBadge

### Issues Resolved

### Check List

### DIFF
--- a/dashboards-observability/public/components/explorer/log_explorer.tsx
+++ b/dashboards-observability/public/components/explorer/log_explorer.tsx
@@ -47,12 +47,13 @@ export const LogExplorer = ({
   const queryRef = useRef();
   const tabIdsRef = useRef();
   const explorerDataRef = useRef();
+  const curSelectedTabIdRef = useRef();
   queryRef.current = queries;
   tabIdsRef.current = tabIds;
   explorerDataRef.current = explorerData;
+  curSelectedTabIdRef.current = curSelectedTabId;
 
   const [tabCreatedTypes, setTabCreatedTypes] = useState({});
-
 
   // Append add-new-tab link to the end of the tab list, and remove it once tabs state changes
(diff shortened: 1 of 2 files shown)


dashboards-observability/public/components/explorer/log_explorer.tsx
```diff
@@ -78,8 +79,9 @@ export const LogExplorer = ({
     }
 
     const index: number = tabIds.indexOf(TabIdToBeClosed);
+    const curSelectedTab = curSelectedTabIdRef.current;
     let newIdToFocus = '';
-    if (TabIdToBeClosed === curSelectedTabId) {
+    if (TabIdToBeClosed === curSelectedTab) {
```
joshuali925, line 84: question, could you explain a bit why curSelectedTabId and its ref are different here?

joshuali925 (approved):

kavithacm (approved):

dashboards-observability/public/components/explorer/log_explorer.tsx
mengweieric: The handTabClose was referring to the stale tabId when it was attached to the UI. When adding the new tab, the current selected tabId would always set to id of the newly added tab, and when you click on any tab before it, the clicked tab data is passed to the handleTabClick handler, and setting the redux store correctly, but since previously the handTabClose is still referring to the stale curSelectedTabId instead of the updated select tabId, it causes this error.

joshuali925: Thanks, got it. I forgot that the tabs are memorized, so this should also be fixed if `curSelectedTabId` is added as a dependency of `memorizedTabs`

PR #354, opened 2021-12-17
Incremented version to 1.2.3.

dblock:
Coming from https://github.com/opensearch-project/opensearch-build/issues/1365.


ps48 (approved):
Thanks for the bump!

mengweieric (approved):

PR #379, opened 2022-01-21
Connected Application Analytics

### LINKED ISSUE #376: Fetch all available applications

eugenesk24:
### Description
This PR includes:
- Remove Availability Level from Application Analytics for the time being
- enable fetching applications
- enable creating applications
- enable renaming applications
- add selection ability to home page table
- cancel button on create page
- remove dummy data
- update mappings to reflect any new additions (add application type to database)

### Issues Resolved
#376 
#323 
#327 
#325 
#340 
#378 

### Check List


dashboards-observability/public/components/application_analytics/components/app_table.tsx
```diff
@@ -27,28 +28,64 @@ import {
 } from '@elastic/eui';
 import _ from 'lodash';
 import React, { ReactElement, useEffect, useState } from 'react';
-import { AppAnalyticsComponentDeps, ApplicationType } from '../home';
+import { AppAnalyticsComponentDeps } from '../home';
+import { getCustomModal } from '../helpers/modal_containers';
 import { pageStyles } from '../../../../common/constants/shared';
+import { ApplicationListType } from 'common/constants/application_analytics';
```
joshuali925, line 34: not sure if this will work, should it be `../../../../common`?

dashboards-observability/public/components/application_analytics/helpers/modal_containers.tsx
```diff
@@ -37,4 +38,34 @@ export const getClearModal = (
       </EuiConfirmModal>
     </EuiOverlayMask>
   );
+};
+
+export const getCustomModal = (
```
joshuali925, line 43: is it possible to reuse the `getCustomModal` from other places? (i think custom_panels has one)

dashboards-observability/public/components/application_analytics/helpers/utils.tsx
```diff
@@ -0,0 +1,9 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+// Name validation 0>Name<=50
+export const isNameValid = (name: string) => {
+  return name.length >= 50 || name.length === 0 ? false : true;
```
joshuali925, line 8: it might be better to return an array of error messages to tell user why it's not valid, in case we add  conditions other than string length like "allowed characters" in the future

joshuali925 (approved):
also need to add UT and IT in backend for the new application type (separate PRs should be fine)

opensearch-observability/src/main/kotlin/org/opensearch/observability/index/ObservabilityIndex.kt
```diff
@@ -108,6 +111,29 @@ internal object ObservabilityIndex {
...
     }
 
+
+    /**
+     * Check if the index mappings have changed and if they have, update them
+     */
+    private fun updateMappings() {
+        val classLoader = ObservabilityIndex::class.java.classLoader
+        val indexMappingSource = classLoader.getResource(OBSERVABILITY_MAPPING_FILE_NAME)?.readText()!!
+        val request = PutMappingRequest(INDEX_NAME)
```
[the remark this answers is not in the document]
eugenesk24: Ah, I see that it's called every time. Is there a reason we call createIndex every time instead of once in the beginning?

dashboards-observability/public/components/application_analytics/components/app_table.tsx
eugenesk24: Oh yea, weird that it didn't complain. I'll add it to be safe!

dashboards-observability/public/components/application_analytics/helpers/modal_containers.tsx
eugenesk24: Yes, but I saw that custom panels and notebooks each have their own so I thought it wouldn't hurt to have one for application analytics too. Do you think I should just re-use one of those still?

opensearch-observability/src/main/kotlin/org/opensearch/observability/index/ObservabilityIndex.kt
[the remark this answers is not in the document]
joshuali925: I asked AK before (he wrote notifications which is where this pattern came from) and he said every request should check this. Don't remember why but my guess is for robustness, like if .opensearch-observability is accidentally deleted. Also checking index exists is a simple read and should have lower overhead than writing to index mappings

dashboards-observability/public/components/application_analytics/helpers/modal_containers.tsx
joshuali925: that was probably by accident, notebooks was merged in without changes while custom panels already added this function by that time. But i think it's better to reduce duplicated code

eugenesk24: On it!

opensearch-observability/src/main/kotlin/org/opensearch/observability/index/ObservabilityIndex.kt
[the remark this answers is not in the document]
eugenesk24: Ah okay, then I'll leave that as a todo item

dashboards-observability/public/components/application_analytics/helpers/utils.tsx
eugenesk24: Got it!

dashboards-observability/common/constants/application_analytics.ts
```diff
@@ -14,6 +14,23 @@ export const TAB_TRACE_TITLE = 'Traces & Spans';
...
+  dateModified: string;
+};
+
+export type ApplicationType = {
+  name: string;
+  description: string;
+  query: string;
+  selectedServices: Array<string>;
+  selectedTraces: Array<string>;
+}
```
ps48, line 36: We should move the types [here](https://github.com/opensearch-project/observability/tree/main/dashboards-observability/common/types)

opensearch-observability/src/main/kotlin/org/opensearch/observability/index/ObservabilityIndex.kt
```diff
@@ -108,6 +113,28 @@ internal object ObservabilityIndex {
         }
     }
 
+    /**
+     * Check if the index mappings have changed and if they have, update them
+     */
+    private fun updateMappings() {
```
ps48, line 119: Why do we need to update mappings? I might be missing something here.

eugenesk24: Since users will already have their indices with the old mapping that doesn't include the Application object type, we need to update their index to reflect the new mapping that does include the Application object type.

dashboards-observability/common/constants/application_analytics.ts
eugenesk24: Got it!

opensearch-observability/src/main/kotlin/org/opensearch/observability/index/ObservabilityIndex.kt
```diff
@@ -84,7 +86,10 @@ internal object ObservabilityIndex {
      */
     @Suppress("TooGenericExceptionCaught")
     private fun createIndex() {
-        if (!isIndexExists(INDEX_NAME)) {
+        if (isIndexExists(INDEX_NAME)) {
+            // TODO: Only update mappings when they have changed so it doesn't run on every request
+            updateMappings()
```
ps48, line 91: Should this change happen when users use the updated observability plugin for the first time. Similar to what we have for old notebooks index. You can check when you look at this in detail.

ps48 (approved):

opensearch-observability/src/main/kotlin/org/opensearch/observability/index/ObservabilityIndex.kt
eugenesk24: You're right, I'll take a look at how the old notebooks index does it and add it in a future PR! 😄

joshuali925 (approved):

PR #413, opened 2022-01-28
Associate panel to application

### LINKED ISSUE #409: Add appId field to custom panels object

eugenesk24:
### Description
Creates a panel upon application creation that includes a new field: applicationId.
Also adds panelId field to application and applicationId to savedVisualization.
Checks if name is empty string.
Minor typo fixes.

### Issues Resolved
#409 
#410 
#411 
#412 

### Check List

### DIFF
--- a/dashboards-observability/public/components/application_analytics/helpers/utils.tsx
+++ b/dashboards-observability/public/components/application_analytics/helpers/utils.tsx
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-// Name validation 0>Name<=50
+// Name validation 
 export const isNameValid = (name: string) => {
   let toast: string[]  = [];
   if (name.length >= 50) {
@@ -11,6 +11,9 @@ export const isNameValid = (name: string) => {
   }
   if (name.length === 0) {
     toast.push('Name must be at least 1 character');
+  }
+  if (name.trim().length === 0) {
+    toast.push('Name must have characters')
   }
   return toast;
 };
--- a/dashboards-observability/public/components/application_analytics/helpers/utils.tsx
+++ b/dashboards-observability/public/components/application_analytics/helpers/utils.tsx
@@ -7,13 +7,10 @@
 export const isNameValid = (name: string) => {
   let toast: string[]  = [];
   if (name.length >= 50) {
-    toast.push('Name must be less than 50 characters');
-  }
-  if (name.length === 0) {
-    toast.push('Name must be at least 1 character');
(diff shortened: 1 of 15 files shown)


dashboards-observability/public/components/application_analytics/helpers/utils.tsx
```diff
@@ -12,5 +12,8 @@ export const isNameValid = (name: string) => {
   if (name.length === 0) {
     toast.push('Name must be at least 1 character');
   }
+  if (name.trim().length === 0) {
+    toast.push('Name must have characters')
```
joshuali925, line 16: I think you can remove the second `if` since it's also covered here, and message can be `Name must not be empty.` to be more consistent with other plugins?

dashboards-observability/server/adaptors/custom_panels/custom_panel_adaptor.ts
```diff
@@ -119,6 +119,7 @@ export class CustomPanelsAdaptor {
         query: '',
         language: 'ppl',
       },
+      applicationId: appId || '',
```
ps48, line 122: Are we keeping this field optional? or the value of the field optional? If the latter, then all the new-panels that are created from panels-home will have an applicationId field which is empty.

dashboards-observability/public/components/custom_panels/home.tsx
```diff
@@ -176,7 +176,7 @@ export const Home = ({ http, chrome, parentBreadcrumb, pplService, renderProps }
       });
```
ps48, line 176: Should we add a filter in panels/home.tsx `in fetchCustomPanels function` , which will filter out the panels which are part of applicationID associated with them?

eugenesk24: Yes! I will be adding that in the next PR, didn't want to make it too long.

dashboards-observability/server/adaptors/custom_panels/custom_panel_adaptor.ts
eugenesk24: I was thinking of using both. Panels that have already been created will not have the field at all, which I will account for when fetching panels and filtering out the ones from app analytics. The new panels will have an empty string for the applicationId. But I guess it is a bit more consistent to not have the field at all for operational panels created outside of app analytics. Typescript makes it look ugly but let me change it to the field being optional.

joshuali925 (approved):

mengweieric (approved):

PR #421, opened 2022-02-01
Support lazy scroll and auto complete for PPL parse command

joshuali925:
### Description
- if `parse` command is present, insert ` | sort - ${timeField} | head 10000` to the query so that `parse` will always process 10k most recent documents regardless of `query.size_limit` setting
- implement lazy scrolling for event explorer
- add `parse` command to autocomplete

Note: `tableRows` (raw documents) and `QueriedTableRows` (aggregated results) share the same scroll limit since currently they cannot coexist in UI

linking parse command support in SQL: https://github.com/opensearch-project/sql/issues/359

### Issues Resolved

### Check List


dashboards-observability/public/components/explorer/data_grid.tsx
```diff
@@ -22,20 +23,49 @@ export function DataGrid(props: DataGridProps) {
...
+    if (!loader.current || limit >= rows.length + PAGE_SIZE) return;
+
+    const observer = new IntersectionObserver(handleObserver, {
+      root: null,
+      rootMargin: '500px',
+      threshold: 0,
+    });
+    observer.observe(loader.current);
+
+    return () => observer.disconnect();
```
eugenesk24, line 44: Question: Is this just the same thing as 
```
observer.disconnect();
return;
```

joshuali925: no, the returned function in useEffect is used for clean up, which will be called when component unmounts. with `observer.disconnect(); return;` then observer is disconnected immediately and won't have any effect

eugenesk24 (approved):
Looks coolio

ps48 (approved):

mengweieric (approved):

PR #475, opened 2022-02-16
Fix create/edit page bug

eugenesk24:

### Description
Fix create edit page bug.
`existingAppId` comes in as the string "undefined" not of type undefined.

### Issues Resolved

### Check List

### DIFF
--- a/dashboards-observability/public/components/application_analytics/components/create.tsx
+++ b/dashboards-observability/public/components/application_analytics/components/create.tsx
@@ -64,7 +64,7 @@ export const CreateApp = (props: CreateAppProps) => {
   const [selectedServices, setSelectedServices] = useState<OptionType[]>([]);
   const [selectedTraces, setSelectedTraces] = useState<OptionType[]>([]);
 
-  const editMode = !!existingAppId;
+  const editMode = existingAppId !== 'undefined';
   const [existingApp, setExistingApp] = useState<ApplicationType>({
     name: '',
     description: '',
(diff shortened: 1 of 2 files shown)


davidcui1225 (approved):

mengweieric (approved):

dashboards-observability/public/components/application_analytics/components/create.tsx
```diff
@@ -64,7 +64,7 @@ export const CreateApp = (props: CreateAppProps) => {
   const [selectedServices, setSelectedServices] = useState<OptionType[]>([]);
   const [selectedTraces, setSelectedTraces] = useState<OptionType[]>([]);
 
-  const editMode = !!existingAppId;
+  const editMode = existingAppId !== 'undefined';
```
joshuali925, line 67: if existingAppId is `undefined` (not the string), then editMode is true but it's not actually edit mode? if so maybe `editMode = existingAppId != null && existingAppId !== 'undefined'` to be consistent with variable naming?

eugenesk24: Not exactly, existingAppId comes from `decodeURIComponent(routerProps.match.params.id)` which returns the string `"undefined"` if the parameter does not exist and the id when it does. So there should not be a case where the existingAppId is `undefined`(not the string)

joshuali925: got it. then why `if (editMode && existingAppId)` instead of `if (editMode)`?

PR #480, opened 2022-02-21
Finish autocomplete logic for after where

### LINKED ISSUE #431: Autocomplete start after base query

### LINKED ISSUE #471: Autocomplete in log config does not complete with correct item selection

eugenesk24:
### Description
- Finish autocomplete logic for suggestions after the base query ( source = index )
- Fix to handle query change and search differently depending on if it's from application analytics

### Issues Resolved
Resolves #431 
Fixes #471 

### Check List

### DIFF
--- a/dashboards-observability/common/constants/autocomplete.ts
+++ b/dashboards-observability/common/constants/autocomplete.ts
@@ -4,6 +4,8 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import { BaseItem } from "@algolia/autocomplete-core";
+
 export const firstCommand = [{ label: 'source' }];
 
 export const pipeCommands = [
@@ -43,7 +45,7 @@ export const numberTypes = [
   'unsigned_long',
 ];
 
-export interface AutocompleteItem {
+export interface AutocompleteItem extends BaseItem {
   input: string;
   itemName: string;
   label: string;
--- a/dashboards-observability/common/constants/autocomplete.ts
+++ b/dashboards-observability/common/constants/autocomplete.ts
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { BaseItem } from "@algolia/autocomplete-core";
+import { BaseItem } from '@algolia/autocomplete-core';
 
 export const firstCommand = [{ label: 'source' }];
 
--- a/dashboards-observability/common/constants/autocomplete.ts
+++ b/dashboards-observability/common/constants/autocomplete.ts
@@ -1,8 +1,8 @@
-/* eslint-disable prettier/prettier */
 /*
  * Copyright OpenSearch Contributors
  * SPDX-License-Identifier: Apache-2.0
  */
+/* eslint-disable prettier/prettier */
 
 import { BaseItem } from '@algolia/autocomplete-core';
 
--- a/dashboards-observability/common/constants/autocomplete.ts
+++ b/dashboards-observability/common/constants/autocomplete.ts
@@ -67,78 +67,70 @@ export interface DataItem {
   doc_count: any;
 }
 
-const JUST_SEARCH_REGEX = new RegExp('\\s*search\\s+source\\s*=\\s*(\\S+)');
-const SEARCH_WHERE_REGEX = new RegExp(
-  '\\s*search\\s+source\\s*=\\s*(\\S+)\\s*\\|\\s*where\\s+\\S+\\s*=\\s*\\S+'
-);
-const SEARCH_MATCH_REGEX = new RegExp(
-  '\\s*search\\s+source\\s*=\\s*(\\S+)\\s*\\|\\s*where\\s+match\\(\\S+,\\s*\\S+\\)'
-);
-const JUST_SOURCE_REGEX = new RegExp('\\s*source\\s*=\\s*(\\S+)');
-const SOURCE_WHERE_REGEX = new RegExp(
-  '\\s*source\\s*=\\s*(\\S+)\\s*\\|\\s*where\\s+\\S+\\s*=\\s*\\S+'
-);
-const SOURCE_MATCH_REGEX = new RegExp(
-  '\\s*source\\s*=\\s*(\\S+)\\s*\\|\\s*where\\s+match\\(\\S+,\\s*\\S+\\)'
-);
-export const EMPTY_REGEX = new RegExp('^\\s*\\S*$');
-export const FIELD_AFTER_COMMAND = new RegExp('^\\s*(dedup|eval|rare|top|rename|where\\s+match\\()\\s+\\S*$');
+const JUST_SEARCH_REGEX = /\s*search\s+source\s*=\s*(\S+)/;
+const SEARCH_WHERE_REGEX = /\s*search\s+source\s*=\s*(\S+)\s*\|\s*where\s+\S+\s*=\s*\S+/;
+const SEARCH_MATCH_REGEX = /\s*search\s+source\s*=\s*(\S+)\s*\|\s*where\s+match\(\S+,\s*\S+\)/;
+const JUST_SOURCE_REGEX = /\s*source\s*=\s*(\S+)/;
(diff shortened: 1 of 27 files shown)


dashboards-observability/common/constants/autocomplete.ts
```diff
@@ -1,8 +1,11 @@
+/* eslint-disable prettier/prettier */
```
davidcui1225, line 1: Can this be moved below the copyright notice?

eugenesk24: Most definitely can

eugenesk24: There are a bunch of other files that have it above the copyright notice, should I change those as well?

dashboards-observability/public/components/common/search/autocomplete.test.tsx
```diff
@@ -105,7 +107,7 @@ describe('autocomplete logic', function () {
         itemName: 'source',
       },
     ] as AutocompleteItem[];
-    const suggestion = await getFullSuggestions(input, dslService);
+    const suggestion = await getFullSuggestions('', input, dslService);
```
davidcui1225, line 110: I don't have any context on autocomplete, why have the function parameters changed to take in an empty string base query?

dashboards-observability/public/components/common/search/autocomplete_logic.ts
```diff
@@ -157,6 +186,7 @@ const filterSuggestions = (suggestions: AutocompleteItem[], prefix: string) => {
 
 // Main logic behind autocomplete (Based on most recent inputs)
 export const getFullSuggestions = async (
+  base: string,
```
joshuali925, line 189: how does base change logic of autocomplete?

dashboards-observability/common/constants/autocomplete.ts
davidcui1225: Yes, the copyright header should be at the top of the file 👍

joshuali925: yes, copyright should be at the top

dashboards-observability/public/components/common/search/autocomplete.test.tsx
eugenesk24: These tests are checking if `getFullSuggestions` is working correctly which is a function that isn't called from Application Analytics so the base query will always be an empty string.

eugenesk24: The reason I have to add the `base` parameter that doesn't get used anyway is that the function for getting suggestions is passed as a prop to the autocomplete component and the other function `getSuggestionsAfterSource` does need that parameter.

dashboards-observability/public/components/common/search/autocomplete_logic.ts
eugenesk24: It does not, refer to this [comment](https://github.com/opensearch-project/observability/pull/480#discussion_r811399199)

dashboards-observability/public/components/common/search/autocomplete.test.tsx
eugenesk24: But as I'm writing this I'm realizing that it may be better to have the prop have two different types and just remove the unnecessary parameter from the function.

eugenesk24: Just tried and it is actually very ugly, may just keep dummy parameter

dashboards-observability/common/constants/autocomplete.ts
```diff
@@ -64,37 +67,127 @@ export interface DataItem {
...
-export const EQUAL_AFTER_FIELD = new RegExp('\\s*where\\s*(\\S+)\\s+');
-export const DATA_AFTER_EQUAL = new RegExp('\\s*where\\s*\\S+\\s*=\\s*\\S*');
-export const PIPE_AFTER_WHERE = new RegExp('\\s*where\\s*\\S+\\s*=\\s*\\S+\\s+');
-export const FIELD_AFTER_MATCH = new RegExp('\\s*where\\s*match\\(\\s*\\S*');
-export const COMMA_AFTER_FIELD = new RegExp('\\s*where\\s*match\\(\\s*(\\S+)\\s+');
-export const DATA_AFTER_COMMA = new RegExp('\\s*where\\s*match\\(\\s*\\S+\\s*,\\s*');
-export const CLOSE_AFTER_DATA = new RegExp('\\s*where\\s*match\\(\\s*\\S+\\s*,\\s*\\S+\\s+');
-export const PIPE_AFTER_MATCH = new RegExp(
-  '\\s*where\\s*match\\(\\s*\\S+\\s*,\\s*\\S+\\s*\\S+\\s*\\)\\s*'
+const JUST_SEARCH_REGEX = new RegExp('\\s*search\\s+source\\s*=\\s*(\\S+)');
```
joshuali925, line 70: are capture groups used? If so could you add some comments on what the regex is capturing? for example this seems to capture the index name, although fails for `source = index1, index2`

also just a note when regex doesn't have variables, `var = /\s*search\s+source\s*=\s*(\S+)/` might be cleaner since backslash doesn't have to be escaped

eugenesk24: Capture groups are mostly used for getting the index and fields but there are places where I am not using the capture groups but the lack of their presence seemed to confuse the matching.

dashboards-observability/public/components/common/search/autocomplete_logic.ts
joshuali925: why not trigger it like `getSuggestions(baseQuery + query, dslService);`, and remove `base` in both get suggestions?

dashboards-observability/common/constants/autocomplete.ts
eugenesk24: Did not know about the being able to use `/` around regex to not have to escape!! I will definitely do that instead. Gosh I wish I had known, I had to delete/add all the escaping every time I was debugging on regex101

eugenesk24: I will also be adding regex patterns for matching multiple indices in a later PR

dashboards-observability/public/components/common/search/autocomplete_logic.ts
eugenesk24: I use `query` to fill the suggestions because the `baseQuery` should not be part of the object which populates the search bar when a suggestion is selected.

dashboards-observability/public/components/common/search/autocomplete.test.tsx
davidcui1225: I see, if you are going to keep the dummy parameter can you add a comment detailing why?

dashboards-observability/public/components/common/search/autocomplete_logic.ts
joshuali925: hm i see, i would probably do 
```js
getSuggestions={appLogEvents ? getSuggestionsAfterSource : (base, curr, dslService) => getFullSuggestions(curr, dslService)}
```
but it feels more ugly. yours is ok i think.

Are the logic from `getFullSuggestions` reusable in `getSuggestionsAfterSource`?

joshuali925 (approved):

davidcui1225 (approved):
Thanks for the change :shipit:

mengweieric (approved):

dashboards-observability/public/components/common/search/autocomplete_logic.ts
eugenesk24: It is very similar but the big difference is that `getFullSuggestions` has to get the words one at a time while `getSuggestionsAfterSource` can get a pre-existing query and parse it and continue with the autocomplete. I think we can just use `getSuggestionsAfterSource` instead of `getFullSuggestions` everywhere else later if we also add suggesting for the source.

PR #494, opened 2022-02-23
Live tail - Event analytics 

### LINKED ISSUE #23: [Live Tail] Supporting live tail on the observability plugin
* Migrate `logviewer-kibana-plugin` repo from ODFE to OpenSearch
* Rule on whether LiveTail should be added to Observability

kavithacm:
### Description
Live tail feature is similar to live stream of data logs but addition to this the user is given the option to choose on which interval the data logs to be refreshed automatically. This excludes the manual labor work for refresh and helps the user to view live data for purposes including debugging their system.

### Issues Resolved
#23 

### Check List


dashboards-observability/public/components/explorer/hooks/use_fetch_events.ts
```diff
@@ -126,6 +130,32 @@ export const useFetchEvents = ({
...
+          res.total = res.total + responseRef.current.total;
+          res.size = res.size + responseRef.current.size;
+        }
+        // // send only first 200 data log arrays
+        // if (res.total > 200) {
+        //   res.jsonData = res.jsonData.slice(0,200);
+        //   res.datarows = res.datarows.slice(0,200);
+        //   res.total = 200;
+        //   res.size = 200;
+        // }
```
ps48, line 150: Do we need to keep these comments?

kavithacm: No, i'll remove it

dashboards-observability/common/utils/live_tail.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  if (isEmpty(rawQuery)) return finalQuery;
+    
+  // convert to moment
+  const start = datemath.parse(startTime)?.format(DATE_PICKER_FORMAT);
+  const end = datemath.parse(endTime)?.format(DATE_PICKER_FORMAT);
+  const tokens = rawQuery.replaceAll(PPL_NEWLINE_REGEX, '').match(PPL_INDEX_INSERT_POINT_REGEX);
+  
+  if (isEmpty(tokens)) return finalQuery;
+  finalQuery = `${tokens![1]}=${tokens![2]} | where ${timeField} >= '${start}' and ${timeField} <= '${end}'${tokens![3]} | sort - timestamp`;
```
ps48, line 37: The sort field `timestamp` seems to be hard-coded. I think it should be replaced with `${timeField}`

kavithacm: yeah okay, i'll replace them

dashboards-observability/common/utils/live_tail.ts
```diff
@@ -0,0 +1,40 @@
...
+import { isEmpty } from 'lodash';
+import datemath from '@elastic/datemath';
+import { DATE_PICKER_FORMAT } from '../../common/constants/explorer';
+import {
+  PPL_INDEX_INSERT_POINT_REGEX,
+  PPL_NEWLINE_REGEX
+} from '../../common/constants/shared';
+
+
+export const insertDateRangeAndSortToQuery = ({
```
joshuali925, line 15: why not use the existing function? https://github.com/opensearch-project/observability/blob/e24f9823e9f8947e28f04ae01b29eec8b189cfe7/dashboards-observability/common/utils/query_utils.ts#L28

kavithacm: existing function didn’t have sort function in it. adding it later was causing an issue so the new function

joshuali925: what was the issue? i think at least call the function in yours to reuse the logic, it could cause more issues if we need to always sync implementations in both functions

kavithacm: I did try to use the same function by modifying the final query before and even after but nothing seemed to work. the sort function is to be included when final query is composed for live tail, hence the new function

joshuali925 (dismissed):

dashboards-observability/common/utils/live_tail.ts
ps48: @kavithacm Josh removed the parse function form the `preprocessQuery` function. Now you can just add another optional boolean argument `liveTail` and check of if you want to add the sort or not at the end of the final query.

dashboards-observability/common/utils/query_utils.ts
```diff
@@ -39,16 +41,19 @@ export const preprocessQuery = ({
...
+  if (isLiveQuery){
+    finalQuery = `${tokens![1]}=${tokens![2]} | where ${timeField} >= '${start}' and ${timeField} <= '${end}'${tokens![3]} | sort - ${timeField}`;
+  } else {
+    finalQuery = `${tokens![1]}=${
+      tokens![2]
+    } | where ${timeField} >= '${start}' and ${timeField} <= '${end}'${
+      tokens![3]
+    }`;
+  };
+  return finalQuery;
```
joshuali925, line 57: i don't think you need to rewrite the part to insert timestamp filter?
```suggestion
  finalQuery = `${tokens![1]}=${
    tokens![2]
  } | where ${timeField} >= '${start}' and ${timeField} <= '${end}'${
    tokens![3]
  }`;
  if (isLiveQuery){
    finalQuery = finalQuery + ' | sort - ${timeField}`;
  }
  return finalQuery; 
```

kavithacm: this works too!

ps48 (approved):
Thanks for the quick changes!

joshuali925 (approved):

kavithacm:
> Thanks for the quick changes!

Thanks for valuable reviews too!!

PR #603, opened 2022-03-23
Modularize and cleanup panel

eugenesk24:
### Description
- Clean up breadcrumbs
- Remove switchToEditViz becuase it is called in home
- Add onEditClick prop that takes the function to call when edit visualization is clicked
- Make add visualization popover a component
- Change button text to sentence case for event analytics to match rest of Observability
- Add cypress tests to check
  - App panels don't show up in Operational Panels table
  - App visualizations don't show up in Event Analytics table
  - App visualizations don't show up in add existing visualization flyout in Operational Panels
  - Show all visualizations of this application in availability configuration
  - Empty panel view has working add button in Operational Panels
  - Empty panel view has working add button in App Analytics

### Issues Resolved

### Check List

### DIFF
--- a/dashboards-observability/.cypress/integration/app_analytics.spec.js
+++ b/dashboards-observability/.cypress/integration/app_analytics.spec.js
@@ -21,10 +21,14 @@ import {
   trace_one,
   trace_two,
   trace_three,
-  spanQueryPartOne,
-  spanQueryPartTwo,
-  spanQueryPartThree,
-  visName,
+  spanQueryOnePartOne,
+  spanQueryOnePartTwo,
+  spanQueryOnePartThree,
+  spanQueryTwoPartOne,
+  spanQueryTwoPartTwo,
+  spanQueryTwoPartThree,
+  visOneName,
+  visTwoName,
   composition,
   newName,
 } from '../utils/app_constants';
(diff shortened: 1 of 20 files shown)


dashboards-observability/.cypress/integration/app_analytics.spec.js
```diff
@@ -22,10 +22,14 @@ import {
...
-  spanQueryPartOne,
-  spanQueryPartTwo,
-  spanQueryPartThree,
-  visName,
+  spanQueryOnePartOne,
+  spanQueryOnePartTwo,
+  spanQueryOnePartThree,
+  spanQueryTwoPartOne,
+  spanQueryTwoPartTwo,
+  spanQueryTwoPartThree,
```
joshuali925, line 30: np but might be better to loop over array, also why does this need to be split into 3 parts?

dashboards-observability/.cypress/integration/app_analytics.spec.js
```diff
@@ -296,36 +315,38 @@ describe('Viewing application', () => {
...
-  it('Adds availability level to visualization', () => {
+  it('Adds availability level to visualization #1', () => {
     cy.get('.euiTab').contains('Panel').click();
     cy.wait(delay);
     cy.get('[aria-label="actionMenuButton"]').click();
     cy.get('.euiContextMenuItem').contains('Edit').click();
     supressResizeObserverIssue();
     cy.wait(delay);
-    cy.get('[data-test-subj="superDatePickerShowDatesButton"]').should('contain', 'Last 24 hours');
+    cy.get('[data-test-subj="superDatePickerShowDatesButton"]').should('contain', 'Last 24 months');
```
joshuali925, line 349: will this test work when it's 2024?

joshuali925 (approved):

dashboards-observability/.cypress/integration/app_analytics.spec.js
eugenesk24: The year 2024??

mengweieric (approved):

davidcui1225 (approved):

dashboards-observability/.cypress/integration/app_analytics.spec.js
eugenesk24: Do the sample logs not get updated every time the cypress tests run??

joshuali925: if it's using dashboards sample data then should be ok, if using the trace data then no

dashboards-observability/.cypress/integration/app_analytics.spec.js
joshuali925: i see, this is ok but you can also just pass the whole string and use a separate function to split by fixed length and feed to cypress.type()

eugenesk24: Yea the only problem is cypress would repeatedly miss certain characters so I had to split the string at those places not at a fixed length

dashboards-observability/.cypress/integration/app_analytics.spec.js
eugenesk24: Most of the tests with time range changes are using dashboards sample data with logs so I think it should be okay.

PR #913, opened 2022-08-03
stats: auto chart added

### LINKED ISSUE #835: [FEATURE] Stats
As per  #787, issue includes below tasks
Stats
- Common Panel Options
- Data configurations: Data configurations section for stat works the same as gauge’s. Each metric is essentially a stat and there will be no breakdowns section for Stat.
- Graph styles
   * Orientation
      - auto(horizontal)
      -  horizontal
      - vertical
   * Legend placement
   * chat mode
      - auto
      - horizontal
      - text mode
   * Title size
   * Value size
   * Series units
   * Precision value for Series
   * Text mode : Control if name and value is displayed or just name.
      - Auto
      - Value and names
      - Values
      - Names
- Thresholds
   - color based thresholds as per reference shared

ramneet-persistent:
### [Feature] Stats chart
- only auto chart type mode is integrated (will raise separate PR for horizontal/text mode chart)
- two sync with query manager

### Issues Resolved
#835 
#1013


spattnaik (changes requested):

dashboards-observability/public/components/visualizations/charts/stats/stats.tsx
```diff
@@ -129,10 +127,11 @@ export const Stats = ({ visualizations, layout, config }: any) => {
     if (metricLength === 1) {
       return 0.5;
     } else {
+      // calculate center of respective axis in each subplot based on metric length
       if (index === 0) {
-        return ((1 / metricLength) * 1) / 2;
+        return 1 / metricLength / 2;
```
mengweieric, line 132: Could you add a little bit more elaborations on your algorithms for calculting center of respective axis in each subplot? May help others to quickly grab the way it does the calculation. thanks

dashboards-observability/public/components/event_analytics/utils/utils.tsx
```diff
@@ -359,3 +364,13 @@ export const filterDataConfigParameter = (parameter: ConfigListEntry[]) =>
   parameter.filter((configItem: ConfigListEntry) => configItem.label);
 
 export const getRoundOf = (value: number, places: number) => value.toFixed(places);
+
+export const getTooltipHoverInfo = ({ tooltipMode, tooltipText }: GetTooltipHoverInfoType) => {
+  if (tooltipMode === 'hidden') {
+    return 'none';
+  } else if (tooltipText === undefined) {
+    return 'all';
+  } else {
```
mengweieric, line 373: similar here, it may not need this else block, just a `return tooltipText;` is enough

dashboards-observability/public/components/visualizations/charts/stats/stats.tsx
```diff
@@ -128,10 +139,7 @@ export const Stats = ({ visualizations, layout, config }: any) => {
           {
             ...STATS_ANNOTATION,
             x: 0 + ANNOTATION_MARGIN_LEFT,
-            y:
-              index > 0
-                ? (index + 1) / metricsLength
-                : 1 / metricsLength,
+            y: index > 0 ? (index + 1) / metricsLength : 1 / metricsLength,
```
mengweieric, line 150: I saw we have more than one similar calculations here like `(index + 1) / metricsLength : 1 / metricsLength` could we extract all similar logics for this type of calculation to a function for all places that need this or similar calculations?

dashboards-observability/public/components/visualizations/charts/stats/stats.tsx
```diff
@@ -0,0 +1,317 @@
...
+            name: metric.label,
+            ...(metricIndex > 0 && {
+              xaxis: `x${metricIndex + 1}`,
+              yaxis: `y${metricIndex + 1}`,
+            }),
+          };
+        });
+
+        // label/text
+        let textLabelTrace: any = {
```
mengweieric: I didn't see textLabelTrace is actually changed later we can use const here.

dashboards-observability/public/components/visualizations/charts/stats/stats.tsx
```diff
@@ -0,0 +1,317 @@
...
+          text: [],
+          type: 'scatter',
+          xaxis: `x${metricsLength + 1}`,
+          yaxis: `y${metricsLength + 1}`,
+          textfont: {
+            size: textSize,
+            color: textColor,
+          },
+        };
+        let reaptedAxis = metricsLength === 1 || orientation === 'auto' ? 'x' : 'y';
```
mengweieric: similar here, could we scan through the file to see if a variable can use const instead of let?

dashboards-observability/public/components/visualizations/charts/stats/stats.tsx
```diff
@@ -0,0 +1,317 @@
...
+        let singleAxis = metricsLength === 1 || orientation === 'auto' ? 'y' : 'x';
```
mengweieric: Please use const.

ramneet-persistent: resolved in next commit

dashboards-observability/public/components/visualizations/charts/stats/stats.tsx
ramneet-persistent: resolved in next commits

dashboards-observability/public/components/visualizations/charts/stats/stats.tsx
ramneet-persistent: updated in next commit

dashboards-observability/public/components/visualizations/charts/stats/stats.tsx
ramneet-persistent: replaced with const as this one is used in this single function only. although revised similar patterns on other code blocks also

dashboards-observability/public/components/event_analytics/explorer/visualizations/config_panel/config_panes/config_controls/config_text_input.tsx
```diff
@@ -0,0 +1,49 @@
...
+export const TextInputFieldItem: React.FC<InputFieldProps> = ({
+  name,
+  title,
+  currentValue,
+  handleInputChange,
+}) => {
+  const [fieldValue, setFieldValue] = useState<string>(currentValue);
+
+  useEffect(() => {
+    setFieldValue('');
```
mengweieric, line 25: The logic here is a little bit confused to me, seems like you always set it to be '' even when there's value for currentValue, and can we maybe name it with a different name as currentName I feel not really a fit here for this component, not sure if it makes sense to you.

mengweieric (approved):

sejli (approved):

PR #915, opened 2022-08-03
Query Manager

### LINKED ISSUE #733: [FEATURE] PPL query re-writer in Dashboards plugin
**Is your feature request related to a problem?**
 We need the UI widgets to re-write and append/extend the PPL stats and other grammar to issue new queries to render visualization changes

**What solution would you like?**
We need a library that can do the re-write operations, preferable with an syntax tree

eg.
https://blog.dangl.me/archive/creating-antlr-applications-in-typescript/

mengweieric:
### Description
Client-side, PPL-grammar-based query manager, including:
- query parser
- query builder

The tests are going with a following PR. Please review it starting from 4e46b74.

### Issues Resolved
#733

### Check List


dashboards-observability/public/components/event_analytics/explorer/explorer.tsx
```diff
@@ -806,27 +807,73 @@ export const Explorer = ({
     );
   };
 
+  const testAntlr = (query: string) => {
+    const qm = new QueryManager();
+    // const pplQueryBuilder = new PPLQueryBuilder();
```
ps48, line 812: Can remove this comment.

dashboards-observability/query_manager/ast/builder/query_builder.ts
```diff
@@ -0,0 +1,8 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export interface QueryBuilder<T> {
+  build: () => T;
+}
```
ps48, line 8: We can move this file to constant.

dashboards-observability/query_manager/ast/types/stats.ts
```diff
@@ -0,0 +1,74 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * stats chunck types
+ */
+export interface StatsAggregationChunk {
+  function: StatsAggregationFunctionChunk;
+  function_alias: string;
+}
```
ps48, line 12: We can move this file to constants as well.

dashboards-observability/query_manager/grammar/OpenSearchPPLLexer.g4
```diff
@@ -0,0 +1,333 @@
...
+
+lexer grammar OpenSearchPPLLexer;
+
+channels { WHITESPACE, ERRORCHANNEL }
+
+
+// COMMAND KEYWORDS
+SEARCH:                             'SEARCH';
+FROM:                               'FROM';
+WHERE:                              'WHERE';
```
ps48, line 15: We can move this file to constants.

joshuali925 (commented):
can the `antlr/bin` dir be git ignored and generated at build time?

dashboards-observability/public/components/event_analytics/explorer/explorer.tsx
mengweieric: these are old, outdated commits, which have been already removed in later commits.

dashboards-observability/query_manager/ast/builder/query_builder.ts
mengweieric: For this and later similar comments, I'm thinking to keep all these in query manager only as they are all specific file/types for query manager, which will be sharing not only by observability plugin.

dashboards-observability/query_manager/ast/types/stats.ts
mengweieric: Please refer to above comments.

dashboards-observability/query_manager/grammar/OpenSearchPPLLexer.g4
mengweieric: similar here as this is the grammar file specific to query manager so i'm thinking to keep these files in query manager only since later this query manager may not be limited to observability.

mengweieric:
> 

Good question, yes we can and maybe also should, but by doing this, do we need to also include the commands generating bin files into starting scripts?

joshuali925:
> Good question, yes we can and maybe also should, but by doing this, do we need to also include the commands generating bin files into starting/build scripts?

what's the command? you can probably add a `postinstall` to package.json

mengweieric:
> > Good question, yes we can and maybe also should, but by doing this, do we need to also include the commands generating bin files into starting/build scripts?
> 
> what's the command? you can probably add a `postinstall` to package.json

https://github.com/opensearch-project/observability/blob/6712046b9c6fd2df8e6215b9af21b50f7becf19e/dashboards-observability/package.json#L13

yes, postinstall is what I should use.

dashboards-observability/query_manager/grammar/OpenSearchPPLLexer.g4
ps48: Got it I didn't know this was also being shared by other plugins later. This makes sense.

joshuali925 (dismissed):

anirudha (approved):

joshuali925 (approved):

Issue #895, opened 2022-07-27
[BUG]: Naming convention for charts field are not common for all charts


nidhisinghai:
@deepaknevdepsl 

**What is the bug?**
Naming convention for charts field are not common for all charts

**How can one reproduce the bug?**

1. Go to Observability/Event Analytics/Explorer page
2. Go to Visualizations
3.  Here for different chart the naming convention is not constant

<img width="176" alt="image" src="https://user-images.githubusercontent.com/103416937/181188908-682c99ae-15c5-46b4-81ee-4212967cacee.png">

<img width="179" alt="image" src="https://user-images.githubusercontent.com/103416937/181189066-a902bd8d-e617-4185-8f6d-f7b4ed06ea17.png">

<img width="179" alt="image" src="https://user-images.githubusercontent.com/103416937/181189274-f87ffe0b-b1a5-4e0b-8395-04afc9a09432.png">

<img width="180" alt="image" src="https://user-images.githubusercontent.com/103416937/181189541-883052c3-d33c-44ad-acde-1b2612bcf083.png">
<img width="170" alt="image" src="https://user-images.githubusercontent.com/103416937/181190217-2b177857-5de1-4670-8d26-a27f39a5a06b.png">

<img width="155" alt="image" src="https://user-images.githubusercontent.com/103416937/181190857-66ec8c9b-f08d-4a4e-9aba-fbcd91e53eca.png">

**What is the expected behavior?**
All the label should follow the same naming convention 
i.e.  The first letter of second word is capital or small.

It should be constant for all the charts

Thanks.
Nidhi Singhai

@abastwar @spattnaik

nidhisinghai:
closing this as correction done

Issue #892, opened 2022-07-22
[BUG] : In Gauge Chart, number of gauges in Data field can be set to negative value.


nidhisinghai:
**What is the bug?**
In Gauge Chart, number of gauges in Data field can be set to negative value.

**How can one reproduce the bug?**
1. Go to Observability/Event Analytics/Explorer page
2. Select Gauge chart from the chart selector
3. In Data field on right side, we can set the number of gauges to negative value

**What is the expected behavior?**
The number of gauges can not be set to negative value

**Do you have any screenshots?**
<img width="166" alt="image" src="https://user-images.githubusercontent.com/103416937/180392860-900ec0f3-7ab0-4258-a1f1-d69353281615.png">

@anirudha @mengweieric 
cc: @abasatwar @spattnaik

nidhisinghai:
closing this as issue solved

Issue #887, opened 2022-07-20
[BUG]: Add update chart button for data configuration panel


ruchika-narang:
**What is the bug?**
Update chart button needs to be added back for data configuration panel because of technical constraints

**What is the expected behavior?**
Update chart button to be visible for data configurations and changes should be applied only when the user clicks on update chart.

ruchika-narang:
Completed.

Issue #863, opened 2022-07-14
Filter Logs View data 


pratibhapandey16:
Filter the data of Logs-View according to the Available fields on the Sidebar

shankha-das:
Hi @mengweieric , @btzeng , @anirudha 

I was working on logs view filter data on the basis of field selection from available fields and pagination. I have a few questions related to it.

1. How many options should be there in 'Rows per page'? (See the screenshot for reference)

2. When we are filtering the logs view data, should we also filter the data inside Accordion or it should display all the data? As per current implementation in Events tab log, when user selects specific fields in the flyout the whole data is visible in tabular form.

3. Please can you confirm on the UI for logs view? Currently it is implemented as in the screenshot.

![MicrosoftTeams-image (13)](https://user-images.githubusercontent.com/93648901/180449145-bcbd304e-b4d6-4616-b34b-0ad52bdc99fd.png)

Thanks,
Shankha

PR #917, opened 2022-08-04
Sprint3 (#62)

abasatwar:
### Description
Visualization implementatation for multiple charts and few bugs are fixed in this PR.

**Issues Resolved**
Filter Logs View data #863
[BUG] : In Gauge Chart, number of gauges in Data field can be set to negative value. #892
Logs view implementation for Visualization #773
[BUG]: Naming convention for charts field are not common for all charts #895
[FEATURE]: Scatter Chart new Config Options Implementation #823
[BUG]: Add update chart button for data configuration panel #887

**Check List**
 

abasatwar:
hi @mengweieric ,
As you requested to earlier , here is list of PR(Internal) where above code is reviewed individually.
https://github.com/abasatwar/observability/pull/61
https://github.com/abasatwar/observability/pull/58
https://github.com/abasatwar/observability/pull/55
https://github.com/abasatwar/observability/pull/56
https://github.com/abasatwar/observability/pull/54
And code is combined to avoid multiple conflicts.
CC: @spattnaik , @anirudha


Thanks,
Amruta

abasatwar:
Hi @mengweieric ,
since DCO check was failing at this PR - https://github.com/opensearch-project/observability/pull/911.
Raised new PR.

Thanks,
Amruta

dashboards-observability/public/components/event_analytics/explorer/visualizations/config_panel/config_panes/config_controls/config_line_chart_styles.tsx
```diff
@@ -33,24 +34,34 @@ export const ConfigLineChartStyles = ({
...
+        case 'markers':
+          return schemas.filter((schema: IConfigPanelOptionSection) =>
+            ['style', 'pointSize'].includes(schema.mapTo)
+          );
+        case 'lines+markers':
+          return schemas.filter(
+            (schema: IConfigPanelOptionSection) => schema.mapTo !== 'interpolation'
+          );
+      }
+    } else {
```
mengweieric, line 57: Maybe further flatten these block like
`
if (vizState?.style) {
} else if (visualizations?.vis?.name === visChartTypes.Scatter) {
} else {}
`

dashboards-observability/public/components/event_analytics/explorer/visualizations/index.tsx
```diff
@@ -73,7 +73,11 @@ export const ExplorerVisualizations = ({
                 handleOverrideTimestamp={handleOverrideTimestamp}
                 handleAddField={(field: IField) => handleAddField(field)}
                 handleRemoveField={(field: IField) => handleRemoveField(field)}
-                isFieldToggleButtonDisabled={true}
+                isFieldToggleButtonDisabled={
```
mengweieric, line 76: What's the end goal we are trying to achieve here? disable toggle buttons when it's log view and there's no viz data? Don't depend on stats keyword in query as this will change soon.

shankha-das: I have changed the logic. My intention was to only enable the Sidebar's field toggle button when the visualization is Logs-View and used the same logic (enabled/disabled) as used in Events tab.

mengweieric: Ok got it, then how about change it to something like
`vis.name !== visChartTypes.LogsView` as I recall that visualization sidebar by default disable all fields.

shankha-das: Yes, along with that I have added this logic which is used in Events Tab: `isEmpty(explorerData.jsonData) || !isEmpty(query[RAW_QUERY].match(PPL_STATS_REGEX))`. Please let me know whether we should keep the next part or we should remove it.

abasatwar:
hi @mengweieric ,
since DCO check was failing , we have raised new PR from this -
https://github.com/opensearch-project/observability/pull/931.
Please merge #931 PR.

Thanks,
Amruta

Issue #912, opened 2022-08-02
[ENHANCEMENT]: Movement of temporary visualization panel data into userconfig


ruchika-narang:
Movement of temporary visualization panel data into userconfig value options for its backward compatibility with saving the visualization for showing in operational panels.

ruchika-narang:
Closing this as it has been merged to main.

PR #929, opened 2022-08-10
[ENHANCEMENT]: Movement of temporary visualization panel data into userconfig

ruchika-narang:
### Description
Movement of temporary visualization panel data into userconfig value options for its backward compatibility with saving the visualization for showing in operational panels.

### Issues Resolved
https://github.com/opensearch-project/observability/issues/912

Note: Internal review has not been done yet.

### Check List
- [ ] New functionality includes movement of visualization panel data to userconfig.

### DIFF
--- a/dashboards-observability/public/components/visualizations/charts/bar/bar.tsx
+++ b/dashboards-observability/public/components/visualizations/charts/bar/bar.tsx
@@ -29,10 +29,12 @@ export const Bar = ({ visualizations, layout, config }: any) => {
   const dataConfigTab =
     visualizations.data?.rawVizData?.bar?.dataConfig &&
     visualizations.data.rawVizData.bar.dataConfig;
-  const xaxis = dataConfigTab?.dimensions
-    ? dataConfigTab.dimensions.filter((item) => item.label)
+  const xaxis = dataConfig?.valueOptions?.dimensions
+    ? dataConfig.valueOptions.dimensions.filter((item) => item.label)
     : [];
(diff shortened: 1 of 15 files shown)


abasatwar (dismissed):
looks good to me

dashboards-observability/public/components/visualizations/charts/bar/bar.tsx
```diff
@@ -68,11 +70,11 @@ export const Bar = ({ visualizations, layout, config }: any) => {
...
   const prepareData = (valueForXSeries) => {
-    return (valueForXSeries.map((dimension: any) => data[dimension.label]))?.reduce(
-      (prev, cur) => {
+    return valueForXSeries
+      .map((dimension: any) => data[dimension.label])
+      ?.reduce((prev, cur) => {
         return prev.map((i, j) => `${i}, ${cur[j]}`);
-      }
-    );
+      });
```
spattnaik, line 77: As this is going to be a longer array with all the values, can't we combine the map and reduce functions? We can fetch the label inside the reduce function instead of having a separate function to loop over all items just to fetch the labels. 

I understand this may not be your code, can we update this to enhance the code?

spattnaik (changes requested):

dashboards-observability/public/components/visualizations/charts/bar/bar.tsx
shankha-das: It has been taken care of in a different branch, hence not modifying the code here. In that branch, we are only using reduce to loop over the data and fetch the label.

mengweieric (approved):

kavithacm (approved):

Issue #965, opened 2022-08-26
[BUG]: Pie chart UI issues


pratibhapandey16:
**What is the bug?**
1. There is a huge space at the top of legends section in chart
2. Values of the pie chart are getting overlapped
3. Mode option for Pie chart has dropdown which has the fields as options 
4. Title over each pie chart with metric sheet was missing

**How can one reproduce the bug?**
1. Go to Observability/Event Analytics/Explorer page
2. Enter the query in PPL query section and select 'Year to date' option from calendar and click on 'Refresh'
5. Navigate to Visualizations page
6. Select Pie chart from chart dropdown
7. Observe the issue

**What is the expected behavior?**
1. There should be a decent space at the top of the legend section
2. Values should not overlap and there should be little space between two pie charts 
3. Mode option for Pie chart should have buttons(Pie and Donut)
4. Title over each pie chart with metric sheet should be present for better clarity between multiple pie charts  

**Do you have any screenshots?**
![image](https://user-images.githubusercontent.com/103417380/186869720-056f9344-7ec9-4f27-870a-42589578342a.png)
![image](https://user-images.githubusercontent.com/103417380/186869795-03744c08-4755-40cf-9882-bbac258e28d3.png)

@anirudha @mengweieric
cc: @abasatwar @spattnaik

PR #974, opened 2022-08-30
Pie chart bug fixes 965

ramneet-persistent:
Resolved [Pie chart bugs](https://github.com/opensearch-project/observability/issues/965)

### DIFF
--- a/dashboards-observability/public/components/event_analytics/utils/utils.tsx
+++ b/dashboards-observability/public/components/event_analytics/utils/utils.tsx
@@ -8,7 +8,7 @@ import { uniqueId } from 'lodash';
 import React from 'react';
 import moment from 'moment';
 import dateMath from '@elastic/datemath';
-import { IExplorerFields, IField } from '../../../../common/types/explorer';
+import { IExplorerFields, IField, ConfigListEntry } from '../../../../common/types/explorer';
 import { DocViewRow, IDocType } from '../explorer/events_views';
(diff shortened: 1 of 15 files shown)


dashboards-observability/public/components/event_analytics/utils/utils.tsx
```diff
@@ -365,11 +365,9 @@ export const filterDataConfigParameter = (parameter: ConfigListEntry[]) =>
 export const getTooltipHoverInfo = ({ tooltipMode, tooltipText }: GetTooltipHoverInfoType) => {
   if (tooltipMode === 'hidden') {
     return 'none';
+  } else if (tooltipText === undefined) {
+    return 'all';
   } else {
-    if (tooltipText === undefined) {
-      return 'all';
-    } else {
```
mengweieric: I think we don't need an else here just `return tooltipText;` as if it matches any branches before it would return already.

mengweieric (approved):

dashboards-observability/public/components/event_analytics/utils/utils.tsx
ramneet-persistent: it is updated in next commit

ps48 (approved):

ramneet-persistent:
combined pr raised with Two way communication for pie chart PR [#1010](https://github.com/opensearch-project/observability/pull/1029), [pull 1029](https://github.com/opensearch-project/observability/pull/1029)

Issue #1105, opened 2022-10-11
[BUG]: If all Series and Dimensions are removed from a query and Refresh button is hit, the app breaks.


DipraAich:
**What is the bug?**
If all Series and Dimensions are removed from a query and Refresh button is hit, the app breaks.

**How can one reproduce the bug?**
1. Add a query with stats and by into the query panel and hit on refresh. Eg: source = opensearch_dashboards_sample_data_logs | stats max(bytes), avg(bytes) by host
2. Once data is loaded update the query and remove all stats and by. Eg: source = opensearch_dashboards_sample_data_logs
3. Hit refresh button.
4. The app crashes.

https://user-images.githubusercontent.com/61182817/195070595-418da412-5602-45b9-b445-408556e4f312.mp4

**What is the expected behavior?**
The app should not crash.

DipraAich:
This issue has been fixed and merged into main. Hence closing this issue.

PR #1110, opened 2022-10-12
[BUG]: If all Series and Dimensions are removed from a query and Refresh button is hit, the app breaks.

DipraAich:
### Description
If all Series and Dimensions are removed from a query and Refresh button is hit, the app breaks.

### Issues Resolved
[#1105](https://github.com/opensearch-project/observability/issues/1105)

### Check List


mengweieric (approved):
LGTM

ps48 (approved):

Issue #907, opened 2022-07-29
[BUG]: Visualizations not appearing in Operational Panels


ruchika-narang:
**What is the bug?**
Saved visualizations does not appear in Operational panel.

**What is the expected behavior?**
Saved visualizations should appear in operational panel after creating the panel

PR #1116, opened 2022-10-13
[BUG]: Visualizations not appearing in Operational Panels fixed

SivaprasadAluri:
### Description
Saved visualizations will appear in operational panel after creating the panel

### Issues Resolved
[BUG]: Visualizations not appearing in Operational Panels (https://github.com/opensearch-project/observability/issues/907)

### DIFF
--- a/dashboards-observability/public/components/custom_panels/helpers/utils.tsx
+++ b/dashboards-observability/public/components/custom_panels/helpers/utils.tsx
@@ -297,7 +297,7 @@ export const displayVisualization = (metaData: any, data: any, type: string) => 
       visualizations={getVizContainerProps({
         vizId: type,
         rawVizData: data,
-        query: { rawQuery: metaData.query },
+        query: { rawQuery: metaData.query === undefined ? {} : metaData.query },
         indexFields: {},
         userConfigs: {},
         explorer: { explorerData: data, explorerFields: data.metadata.fields },
--- a/dashboards-observability/public/components/custom_panels/helpers/utils.tsx
+++ b/dashboards-observability/public/components/custom_panels/helpers/utils.tsx
@@ -299,7 +299,7 @@ export const displayVisualization = (metaData: any, data: any, type: string) => 
         rawVizData: data,
         query: { rawQuery: metaData.query === undefined ? {} : metaData.query },
         indexFields: {},
-        userConfigs: {},
+        userConfigs: metaData.user_configs,
         explorer: { explorerData: data, explorerFields: data.metadata.fields },
       })}
     />
--- a/dashboards-observability/public/components/custom_panels/helpers/utils.tsx
+++ b/dashboards-observability/public/components/custom_panels/helpers/utils.tsx
@@ -297,9 +297,9 @@ export const displayVisualization = (metaData: any, data: any, type: string) => 
       visualizations={getVizContainerProps({
         vizId: type,
         rawVizData: data,
-        query: {},
+        query: { rawQuery: metaData.query },
         indexFields: {},
-        userConfigs: metaData.user_configs,
+        userConfigs: data?.jsonData,
         explorer: { explorerData: data, explorerFields: data.metadata.fields },
       })}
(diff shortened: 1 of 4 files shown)


dashboards-observability/public/components/custom_panels/helpers/utils.tsx
```diff
@@ -299,7 +299,7 @@ export const displayVisualization = (metaData: any, data: any, type: string) =>
         rawVizData: data,
         query: { rawQuery: metaData.query },
         indexFields: {},
-        userConfigs: data?.jsonData,
+        userConfigs: {},
```
mengweieric, line 302: Why are we change it from metaData.user_configs to data?.jsonData and removed it in the end? Did we do BWC?

ps48: +1 @SivaprasadAluri Can you please explain why was this change added?

dashboards-observability/public/components/custom_panels/helpers/utils.tsx
```diff
@@ -297,9 +297,9 @@ export const displayVisualization = (metaData: any, data: any, type: string) =>
       visualizations={getVizContainerProps({
         vizId: type,
         rawVizData: data,
-        query: {},
+        query: { rawQuery: metaData.query === undefined ? {} : metaData.query },
```
ps48, line 300: In query parameter Why are we checking `metaData.query === undefined`? The statement will never be reached while being true. Check: https://github.com/opensearch-project/observability/blob/d243e918aff872efe7fed3769662ebddfec56b2e/dashboards-observability/public/components/custom_panels/helpers/utils.tsx#L292-L294

You may simply change this to:
```
query: { rawQuery: metaData.query }
```

dashboards-observability/test/panels_constants.tsx
```diff
@@ -12,10 +12,23 @@ export const sampleSavedVisualization = {
   visualization: {
     id: 'oiuccXwBYVazWqOO1e06',
     name: 'Flight Count by Origin',
-    query:
-      'source=opensearch_dashboards_sample_data_flights | fields Carrier,FlightDelayMin | stats sum(FlightDelayMin) as delays by Carrier',
+    query: {
+      rawQuery:
```
ps48, line 16: @SivaprasadAluri @mengweieric I don't see this structure change in the backend: https://github.com/opensearch-project/observability/blob/main/opensearch-observability/src/main/kotlin/org/opensearch/observability/model/SavedVisualization.kt#L24-L25

Are we yet to have a change in the backend for this?

dashboards-observability/public/components/custom_panels/helpers/utils.tsx
SivaprasadAluri: @mengweieric,@ps48 sorry i had reverted back to metaData.user_configs but missed the change during the last commit, its been reverted and tested.

dashboards-observability/public/components/custom_panels/helpers/utils.tsx
SivaprasadAluri: @ps48 we do have a test case where, we are checking with empty visualization, so i had to fix it with this. https://github.com/opensearch-project/observability/blob/e4a199be779eac368d0682fc636ee8d12cbaa05b/dashboards-observability/public/components/custom_panels/helpers/__tests__/utils.test.tsx#L128

Please let me know if we have to remove the test case or update the test case with valid visualization data.

dashboards-observability/test/panels_constants.tsx
SivaprasadAluri: @ps48 rawQuery is not needed here, and its been reverted back to query

mengweieric:
This PR is pending BWC support/updates as discussed with persist team.

SivaprasadAluri:
The bug will be fixed in backward compatibility PR, this Pull is no longer needed.

Issue #725, opened 2022-05-06
[FEATURE][Event Analytics] Some feature requests to improve user experience


joshuali925:
## 1. Link some help on how to create a visualization with PPL:

Right now if user runs a query `source = apache_logs`, the documents are displayed in "Events" tab, but "Visualizations" tab shows "No results found". This can be confusing because there could be results, it actually means no aggregated results found. 

<img width="711" alt="image" src="https://user-images.githubusercontent.com/28062824/167223452-f07bd1e8-3e4d-4e9e-b24c-640c547f34cf.png">

**Proposed fix**: would be helpful if there's an explanation that the `stats` command is required for visualizations.

## 2. Need better saving logic when saving queries/visualizations:

<img width="299" alt="image" src="https://user-images.githubusercontent.com/28062824/167223594-216ef6ca-2bb8-420d-a28b-3edacd8c27bc.png">

Current behavior of "Save" will overwrite the existing saved object if it exists. This may or may not be what the user wants, and there is no way to undo the save if user accidentally overwrote their saved object. Also this prevents user from adding multiple visualizations to Operation Panel as every "save" will overwrite previously saved ones. 

**Proposed fix**: it would be more clear if there is a "Save as new" button, or make it rewrite object only if the name did not change.

**Minor issues on UX**: "Name your savings" can be reworded to "Enter a name for the saved query." and "Enter a name for the saved visualization.", and input validation should not trigger (turning to red) until user enters an invalid name.

### 3. Persist user input in session storage or url:

Users lose their inputs in query bars if they navigate to another query tab before running the query. This can happen if they are using other tabs as references while writing the query. Additionally user looses all inputs if they refresh the tab without saving.

**Proposed fix**: Persist input in query bars across tab switch, would be better if it's part of the `URLSearchParams` so user can copy/paste URLs for any query

PR #1178, opened 2022-10-21
Persist redux state in browser

### LINKED ISSUE #1177: [FEATURE] Persist redux data on browser
**Is your feature request related to a problem?**
Currently, redux state data gets lost on page refresh, which leads to bad user experience.

**What solution would you like?**
Propose to use redux persist 6.0 which provides out of box solution for redux toolkit and just need some additional but straitforward configurations. 

**What alternatives have you considered?**
As far as I know there aren't too many options out there for redux toolkit specifically, there is one middleware called redux-storage where its last activity happened roughly 6 years ago. Feel free to left comments if there are other better options available.

**Do you have any additional context?**
The downside of redux persist 6.0 is: when last time there were updates is 12 months ago, and the last released version is roughly 3 years ago as 6.0.0. But currently it works fine in our cases with current functionalities it supports.

mengweieric:
Signed-off-by: Eric Wei <menwe@amazon.com>

### Description
Currently, redux state data gets lost on page refresh, which leads to bad user experience. This change is to address this problem.

### Issues Resolved
#1177 

### Check List

### DIFF
--- a/dashboards-observability/public/framework/redux/store/index.ts
+++ b/dashboards-observability/public/framework/redux/store/index.ts
@@ -5,21 +5,45 @@
 
 import { configureStore } from '@reduxjs/toolkit';
 import rootReducer from '../reducers';
+import storage from 'redux-persist/lib/storage';
+import {
+  persistStore,
+  persistReducer,
+  FLUSH,
+  REHYDRATE,
+  PAUSE,
+  PERSIST,
+  PURGE,
+  REGISTER,
+} from 'redux-persist';
+
+const persistConfig = {
+  key: 'root',
+  storage,
+};
+
+const persistedReducer = persistReducer(persistConfig, rootReducer);
 
 const store = configureStore({
-  reducer: rootReducer,
-  middleware: (getDefaultMiddleware) => getDefaultMiddleware(),
+  reducer: persistedReducer,
+  middleware: (getDefaultMiddleware) =>
+    getDefaultMiddleware({
+      serializableCheck: {
+        ignoredActions: [FLUSH, REHYDRATE, PAUSE, PERSIST, PURGE, REGISTER],
+      },
+    }),
   devTools: process.env.NODE_ENV !== 'production',
   enhancers: [],
 });
 
 if (process.env.NODE_ENV === 'development' && module.hot) {
   module.hot.accept('./rootReducer', () => {
-    const newRootReducer = require('./rootReducer').default;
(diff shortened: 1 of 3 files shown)


dashboards-observability/public/framework/redux/store/index.ts
```diff
@@ -5,21 +5,45 @@
 
 import { configureStore } from '@reduxjs/toolkit';
 import rootReducer from '../reducers';
+import storage from 'redux-persist/lib/storage';
```
joshuali925, line 8: should this use session storage instead of local storage?

joshuali925 (approved):

pjfitzgibbons (changes requested):
Eric, could you please describe in more detail how the user is affected before this change?
What happens if the user navigates away from a page by URL and then back?

dashboards-observability/public/framework/redux/store/index.ts
mengweieric: Yes, we should use session storage, the default storage is localStorage

mengweieric:
> Eric, could you please describe in more detail how the user is affected before this change? What happens if the user navigates away from a page by URL and then back?

Sure, before this change, if some user actions (for example, directly trigger a saved visualizations url in address bar, or simply just reload the page) trigger a page reload/refresh, all the event/visualization data for each tab in event analytics if there's any will get lost. Navigate to other pages from URL or other user actions within the domain would not cause data loss, which is also not part of the problem this change is trying to address.

dashboards-observability/public/framework/redux/store/index.ts
derek-ho, line 8: Have we considered using url parameters? That way users can share a page with given filters or bookmark a page with commonly used filters for convenience?

pjfitzgibbons (approved):

dashboards-observability/public/framework/redux/store/index.ts
mengweieric: Yes, this is a feature, you may create an issue for a feature request.

joshuali925:
linking issue https://github.com/opensearch-project/observability/issues/725 this fixes item 3

derek-ho (dismissed):
LGTM pending session/local storage changes

derek-ho (approved):

anirudha (approved):

Issue #1174, opened 2022-10-21
[BUG]: HeatMap on the update of PPL query configurations should not have values


SivaprasadAluri:
**What is the bug?**
A clear and concise description of the bug.

**How can one reproduce the bug?**

**What is the expected behavior?**

**What is your host/environment?**
 - Plugins

**Do you have any screenshots?**

**Do you have any additional context?**
Add any other context about the problem.

SivaprasadAluri:
@mengweieric, Updated the Code for Heatmap where user quired with not one Series or more or less than two dimensions in chart the message is updated as Invalid Configs, also updated the tooltips according to it, Please let me know if changes are required.
CC: @ruchika-narang 

https://user-images.githubusercontent.com/110654651/198617499-da837430-c06d-42df-8fe1-86446b7c585f.mp4

PR #1220, opened 2022-11-02
[BUG]: Fix for HeatMap configurations should have only one Series and two Dimensions.

SivaprasadAluri:
### Description
Fix for HeatMap configurations should have only one Series and two Dimensions.

### Issues Resolved
[BUG]: HeatMap on the update of PPL query configurations should not have values https://github.com/opensearch-project/observability/issues/1174

### DIFF
--- a/dashboards-observability/common/constants/explorer.ts
+++ b/dashboards-observability/common/constants/explorer.ts
@@ -285,3 +285,8 @@ export const SERIES_POSITION_OPTIONS = [
   { id: htmlIdGenerator('ct')(), label: 'Left', side: 'left' },
   { id: htmlIdGenerator('ct')(), label: 'Right', side: 'right' },
 ];
+
+export const HEATMAP_DATA_CONFIG_HINTS_INFO = {
+  [AGGREGATIONS]: 'only one Series is allowed',
+  [GROUPBY]: 'only two Dimensions are allowed',
+};
--- a/dashboards-observability/common/constants/explorer.ts
+++ b/dashboards-observability/common/constants/explorer.ts
@@ -281,6 +281,10 @@ export const DATA_CONFIG_HINTS_INFO = {
   [BREAKDOWNS]:
     "Defines how each series is broken down. Breakdowns are 'by' clauses that subdivide the existing series.",
 };
+export const SERIES_POSITION_OPTIONS = [
+  { id: htmlIdGenerator('ct')(), label: 'Left', side: 'left' },
+  { id: htmlIdGenerator('ct')(), label: 'Right', side: 'right' },
+];
 
 export const HEATMAP_DATA_CONFIG_HINTS_INFO = {
   [AGGREGATIONS]: `${DATA_CONFIG_HINTS_INFO[AGGREGATIONS]} and max one series is allowed`,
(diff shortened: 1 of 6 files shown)


dashboards-observability/common/constants/explorer.ts
```diff
@@ -280,3 +280,8 @@ export const DATA_CONFIG_HINTS_INFO = {
   [BREAKDOWNS]:
     "Defines how each series is broken down. Breakdowns are 'by' clauses that subdivide the existing series.",
 };
+
+export const HEATMAP_DATA_CONFIG_HINTS_INFO = {
+  [AGGREGATIONS]: 'only one Series is allowed',
```
mengweieric, line 285: Ideally what we can do here is to have shared info and custom field for different visualizations if they have additional/different content to show for info icon. but all visualizations where it uses dimensions/series/breakdowns should use above general text for explanations. for example, if heatmap has such restrictions, then you can concatenate string to form maybe something like `${sharedInfo}. ${heatmapCustomInfo}`.

dashboards-observability/public/components/event_analytics/explorer/visualizations/config_panel/config_panes/config_controls/data_config_panel_fields.tsx
```diff
@@ -44,19 +45,34 @@ export const DataConfigPanelFields = ({
...
+        ? list.length >= 1
+        : isEmpty(time_field)
+        ? list.length >= 2
+        : list.length >= 1;
     // condition for line and scatter for dimensions section.
     return name === GROUPBY && (list.length >= 1 || !isEmpty(time_field));
   };
 
-  const { time_field, unit, interval } = dimensionSpan;
+  const toolTipTextGenerator = (isAgg: boolean) => {
```
mengweieric, line 65: In this way we are missing breakdowns here, and this implementations are limited to only be able to serve text for series/dimensions

dashboards-observability/public/components/visualizations/charts/maps/heatmap.tsx
```diff
@@ -58,16 +66,23 @@ export const HeatMap = ({ visualizations, layout, config }: any) => {
     isEmpty(xaxisField) ||
     isEmpty(yaxisField) ||
     isEmpty(zMetrics) ||
-    isEmpty(queriedVizData[xaxisField.label]) ||
+    isEmpty(queriedVizData[xaxisField.name]) ||
```
mengweieric, line 69: Question: why are we changing from label to name?

anirudha (dismissed):

dashboards-observability/public/components/visualizations/charts/maps/heatmap.tsx
SivaprasadAluri: as we, included time_field(span) into dimensions and id does not contain label as common. in both dimensions and time_field(span)  the common filed was **name**, so we did changed to it.

dashboards-observability/common/constants/explorer.ts
SivaprasadAluri: @mengweieric, updated the constant as below, it will get concatenated with actual series and dimensions info as suggested. and please suggest if changes are needed for text.
```
export const HEATMAP_DATA_CONFIG_HINTS_INFO = {
  [AGGREGATIONS]: `${DATA_CONFIG_HINTS_INFO[AGGREGATIONS]} and max one series is allowed`,
  [GROUPBY]: `${DATA_CONFIG_HINTS_INFO[GROUPBY]} and max two dimensions are allowed`,
};
```

PR #1266, opened 2022-11-08
fix bug by keeping track of interval within the explorer

### LINKED ISSUE #1249: [BUG] Interval for histogram in event analytics is inconsistent
**What is the bug?**
Clicking in the patterns table seems to reset the selected interval, but the selected interval in the dropdown is not changed - this should be consistent, either reset to default and have dropdown change with it or keep the same interval and update the chart accordingly

**How can one reproduce the bug?**
1. Go to event analytics and run a query
2. histogram interval says `minute` but is actually `year`
3. change `minute` to `hour`
4. change `hour` back to `minute`
5. the histogram interval is now `minute`
6. update state (update the query, or click a pattern in the patterns table to add a filter)
7. the histogram interval is back to `year`

**What is the expected behavior?**
The interval should persist

**What is your host/environment?**
 - Plugins

**Do you have any screenshots?**

**Do you have any additional context?**
Add any other context about the problem.

derek-ho:
Signed-off-by: Derek Ho <dxho@amazon.com>

### Description
Keeps track of interval selected in the explorer and timechart header so that the interval is not reset whenever the query changes

### Issues Resolved
#1249 

### Check List


dashboards-observability/public/components/event_analytics/explorer/explorer.tsx
```diff
@@ -389,13 +393,15 @@ export const Explorer = ({
       getVisualizations();
       getAvailableFields(`search source=${curIndex}`);
     } else {
-      findAutoInterval(startTime, endTime);
+      if (!selectedIntervalRef.current || selectedIntervalRef.current.text === 'Auto') {
+        findAutoInterval(startingTime, endingTime);
```
joshuali925, line 397: what's the difference between startingTime and startTime?

dashboards-observability/public/components/event_analytics/explorer/explorer.tsx
```diff
@@ -168,13 +173,11 @@ export const Explorer = ({
   liveTailTabIdRef.current = liveTailTabId;
   liveTailNameRef.current = liveTailName;
 
-  let minInterval = 'y';
   const findAutoInterval = (start: string = '', end: string = '') => {
-    if (start?.length === 0 || end?.length === 0 || start === end)
-      return ['d', [...TIME_INTERVAL_OPTIONS]];
     const momentStart = dateMath.parse(start)!;
     const momentEnd = dateMath.parse(end)!;
```
joshuali925, line 178: could you also change this
```suggestion
    const momentEnd = dateMath.parse(end, { roundUp: true })!;
```

dashboards-observability/public/components/event_analytics/explorer/explorer.tsx
```diff
@@ -678,10 +684,13 @@ export const Explorer = ({
                           <TimechartHeader
                             dateFormat={'MMM D, YYYY @ HH:mm:ss.SSS'}
                             options={timeIntervalOptions}
-                            onChangeInterval={(intrv) => {
+                            onChangeInterval={(selectedIntrv) => {
+                              const intervalOptionsIndex = timeIntervalOptions.findIndex(item => item.value === selectedIntrv)
+                              const intrv = selectedIntrv.length > 2 ? selectedIntrv.slice(5) : selectedIntrv
```
joshuali925, line 689: `timeIntervalOptions` values are one character, when would `selectedIntrv.length > 2`?

dashboards-observability/public/components/event_analytics/explorer/explorer.tsx
```diff
@@ -389,13 +393,15 @@ export const Explorer = ({
...
+      if (!selectedIntervalRef.current || selectedIntervalRef.current.text === 'Auto') {
+        findAutoInterval(startingTime, endingTime);
+      }
       if (isLiveTailOnRef.current) {
         getLiveTail(undefined, getErrorHandler('Error fetching events'));
       } else {
         getEvents(undefined, getErrorHandler('Error fetching events'));
       }
-      getCountVisualizations(minInterval);
+      getCountVisualizations(selectedIntervalRef.current!.value.length > 2 ? selectedIntervalRef.current!.value.slice(5) : selectedIntervalRef.current!.value);
```
joshuali925, line 404: > `    selectedIntervalRef.current = ({ text: 'Auto', value: 'auto_' + minInterval })`

is the `auto_` needed? seems `value` can either be `auto_x` or `x` but why

dashboards-observability/public/components/event_analytics/explorer/explorer.tsx
```diff
@@ -678,10 +684,13 @@ export const Explorer = ({
...
                             options={timeIntervalOptions}
-                            onChangeInterval={(intrv) => {
+                            onChangeInterval={(selectedIntrv) => {
+                              const intervalOptionsIndex = timeIntervalOptions.findIndex(item => item.value === selectedIntrv)
+                              const intrv = selectedIntrv.length > 2 ? selectedIntrv.slice(5) : selectedIntrv
                               getCountVisualizations(intrv);
+                              selectedIntervalRef.current = timeIntervalOptions[intervalOptionsIndex]
                             }}
-                            stateInterval="auto"
+                            stateInterval={selectedIntervalRef.current?.value || ''}
```
joshuali925, line 693: when would `selectedIntervalRef.current?.value` not exist?

dashboards-observability/public/components/event_analytics/explorer/explorer.tsx
derek-ho: It would be > 2 when the selected option is 'Auto', which has value 'auto_X'

dashboards-observability/public/components/event_analytics/explorer/explorer.tsx
derek-ho: I think the main issue I can see here is that if it is just x it would be ambiguous with the other x option - i.e. if Auto is 'Auto', 'w', there may be ambiguity between auto and 'Week' 'w' options

dashboards-observability/public/components/event_analytics/explorer/explorer.tsx
derek-ho: I was thinking it may not exist before we have a chance to call fetch data - but maybe I am overthinking - should I just use a `!` and assert that it exists by the time we render the timechart header?

dashboards-observability/public/components/event_analytics/explorer/explorer.tsx
derek-ho: I may need someone more familiar with this component/code (@mengweieric ?) - but from what I can tell this explorer is rendered by both Application analytics and event analytics - for event analytics startTime is undefined, resulting in auto interval not getting picked up. In either case startingTime should be already set above based on the query so I think(?) this is a more general implementation

dashboards-observability/public/components/event_analytics/explorer/explorer.tsx
derek-ho: Done in : [1776e8c](https://github.com/opensearch-project/observability/pull/1266/commits/1776e8cdbc4e0ca487c6519d80ce731dd8a70cee). @joshuali925 I also did this anywhere else I saw datemath.parse in endtime/lte and also removed duplicate code fragment. Can you double check? Also do we need to do anything for startTime?

dashboards-observability/public/components/event_analytics/explorer/explorer.tsx
joshuali925: i think you can change https://github.com/opensearch-project/observability/blob/af2c6acbf860b5246cbe0a92dc7f9112cdd81377/dashboards-observability/public/components/event_analytics/explorer/timechart_header/timechart_header.tsx#L35
to `stateInterval?: string` and remove the `|| ''` here

dashboards-observability/public/components/event_analytics/explorer/explorer.tsx
joshuali925: true. might be an opinion but can this code be more explicit? e.g. `selectedIntervalRef.current!.value.replace(/^auto_/, '')`

dashboards-observability/public/components/event_analytics/explorer/explorer.tsx
joshuali925: looks good, startTime by default rounds down so it's ok. this is for when user selects 'this week' or 'this month'

dashboards-observability/public/components/event_analytics/explorer/explorer.tsx
derek-ho: Fixed in [25ed4ec](https://github.com/opensearch-project/observability/pull/1266/commits/25ed4ec02ba006512631e22aacf038f9ca599525)

dashboards-observability/public/components/event_analytics/explorer/explorer.tsx
derek-ho: Agreed and fixed in [25ed4ec](https://github.com/opensearch-project/observability/pull/1266/commits/25ed4ec02ba006512631e22aacf038f9ca599525)
[19aaff1](https://github.com/opensearch-project/observability/pull/1266/commits/19aaff1399fd9a979917986361588dd144f0ab93)

joshuali925 (approved):

dashboards-observability/public/components/event_analytics/explorer/explorer.tsx
mengweieric: Yes confirmed. Event analytics previously does not use any passed in start/starting time, I think later there's live tail introduced with starting/ending time for differentiating existing start/end time just for reusing the exiting data fetching

mengweieric (approved):

joshuali925:
CI failed, retrying. not sure why recently bootstrap often fails

joshuali925 (approved):

PR #1352, opened 2022-12-21
Removing explicit jackson dependencies

### LINKED ISSUE #1347: Remove jackson-databind and jackson-annotations dependencies now coming from core
jackson-databind and jackson-annotations were added as api dependencies in `server` of opensearch core.  Therfore, all plugins now have these as transitive dependencies.  Plugins that currently consume these dependencies directly will break while building unless the direct dependencies are removed.  If your plugin does not consume jackson-databind or jackson-annotations directly, feel free to close the related issue in your plugin's repository.

Example of solution in security plugin: https://github.com/opensearch-project/security/pull/2325

ps48:
Signed-off-by: Shenoy Pratik <sgguruda@amazon.com>

### Description
Removing explicit jackson dependencies

### Issues Resolved
#1347 

### Check List

### DIFF
--- a/opensearch-observability/build.gradle
+++ b/opensearch-observability/build.gradle
@@ -23,7 +23,6 @@ buildscript {
         common_utils_version = System.getProperty("common_utils.version", opensearch_build)
         kotlin_version = System.getProperty("kotlin.version", "1.6.0")
         job_scheduler_version = System.getProperty("job_scheduler.version", opensearch_build)
-        jackson_version = "2.14.1"
     }
 
     repositories {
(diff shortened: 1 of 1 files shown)


YANG-DB (approved):

opensearch-observability/build.gradle
```diff
@@ -23,7 +23,6 @@ buildscript {
         common_utils_version = System.getProperty("common_utils.version", opensearch_build)
         kotlin_version = System.getProperty("kotlin.version", "1.6.0")
         job_scheduler_version = System.getProperty("job_scheduler.version", opensearch_build)
-        jackson_version = "2.14.1"
```
derek-ho, line 26: curious as to do we need to add a line like they did in sql : https://github.com/opensearch-project/sql/pull/1169/files?

derek-ho (approved):

opensearch-observability/build.gradle
derek-ho: `apply plugin: 'opensearch.java'`

ps48: I think SQL needs it as they are still using the `versions.jackson` from the OpenSearch config. This may be needed for SQL due to its direct dependency from other gradle libraries. We don't need it as we just use the version provided by the OS-core.

PR #1446, opened 2023-03-04
add catalog meta info in the index template

YANG-DB:
### Description
add catalog meta in the index template
add catalog API for schema query
add index pattern update as part of the Integration-API.md
add asset index patterns exists validation

### Issues Resolved
API documentation update
 - https://github.com/opensearch-project/observability/issues/1461
 - https://github.com/opensearch-project/observability/issues/1462
 - https://github.com/opensearch-project/observability/issues/1467
### Check List
  - [x] New functionality has user manual doc added


joshuali925 (approved):

integrations/docs/Integration-API.md
```diff
@@ -132,7 +132,7 @@ Store API:
 ```
 POST _integration/store/$instance_name
 ```
-The $instance_name represents the specific name the integration was instanciated with - for example an Nginx Integration can be a template for multiple Nginx instances
+The $instance_name represents the specific name the integration was instanced with - for example, Nginx Integration can be a template for multiple Nginx instances
```
Swiddis, line 135: Instantiated/created

integrations/docs/Integration-plugin-tasks.md
```diff
@@ -19,7 +19,29 @@ template mapping for each catalog schema entity. This will allow the future Inte
...
+ - type
+ - catalog
+ - category
+ - version
+
+`GET _integration/catalog?filter=type:Logs&catalog:Observability&category:web`
+
+ This call will result with the appropriate index_template IDs corresponding to the query:
+
+```json
```
Swiddis, line 33: It's worth noting that Github supports [json5](https://json5.org/) language formatting, not important to change now but for future reference setting `json5` will avoid the red syntax errors.

Swiddis (approved):

integrations/docs/Integration-plugin-tasks.md
YANG-DB: cool - thanks

Swiddis (dismissed):

integrations/docs/API/swagger.yaml
```diff
@@ -0,0 +1,287 @@
...
+            schema:
+              $ref: '../../../schema/system/integration-instance.schema'
+      responses:
+        '201':
+          description: item created
+        '400':
+          description: 'invalid input, object invalid'
+        '409':
+          description: an existing item already exists
+  /store/{id}/status:
```
derek-ho, line 183: we will need a way to bulk check status - or else we will need to make one call for each added api to check status - which may be too much

integrations/docs/API/swagger.yaml
```diff
@@ -0,0 +1,287 @@
...
+      responses:
+        '200':
+          description: integration store status
+          content:
+            application/json:
+              schema:
+                $ref: '../../../schema/system/integration-instance.schema'
+        '400':
+          description: bad input parameter
+  /store/{id}/validate:
```
derek-ho, line 207: Is this swagger for both internal/external APIs? This seems like it should be an internal API called only after user tries to store something?

Swiddis: I would also expect validation to be a part of the storage routine. My understanding is that it's because an integration has many referential parts and you can't do a full validation before every part is loaded. But where are we getting circular references that causes that?

integrations/docs/API/swagger.yaml
```diff
@@ -0,0 +1,287 @@
...
+      description: |
+        By passing in the appropriate search attributes, you can search for
+        available integration in the repository (available for loading into the store)
+      parameters:
+        - in: query
+          name: searchString
+          description: pass an optional search string for looking up available Integrations
+          required: false
+          schema:
+            type: string
```
derek-ho, line 86: For this one, I do not see a separate entry in this swagger for /repository/${id}. I think it makes sense to separate it out - for /repository?filter we can return all integrations, but only necessary to show on UI - like name, description, img url, etc.
Then for /repository/${id} we will need to fetch things like SSFO fields, etc.

derek-ho: I was also checking the schema/sample json files - I couldn't find one with SSFO fields listed - only relative path to some schema file. / @YANG-DB can you point out where that would be/ add it if not present? Front end would need to get the actual mapping/fields, or is the plan to expose another API to query the fields?

integrations/docs/API/swagger.yaml
```diff
@@ -0,0 +1,287 @@
...
+            maximum: 50
+      responses:
+        '200':
+          description: search results matching criteria
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '../../../schema/system/integration.schema'
```
derek-ho, line 103: The current figma suggests having a "status" section, which indicates if this integration was added or not. Would it be trivial to add another field on this schema about whether it was added or not?

derek-ho: The UI could also do this by cross-checking with the "added" list, but if it is trivial to add it to the schema/API can we do it please?

Swiddis (approved):

ps48 (approved):

Issue #1443, opened 2023-03-01
[FEATURE] Integration loading workflow


YANG-DB:
[Previous Step](https://github.com/opensearch-project/observability/issues/1441)

**Is your feature request related to a problem?**
### Load Integration
As part of the Integration Ux workflow, once the Integration plugin has loaded and was selected by the user for loading into the system - the B/E should initiate the loading process and display the appropriate status to reflect the loading steps...

This phase follows the [previous step](https://github.com/opensearch-project/observability/issues/1441) in which the user has filtered the Integrations from the repository and selected a specific one to load into the system

### Integration loading state machine
```
 - LOADING
   - VALIDATION
     - UPLOAD
       - READY
```

The API needed from the backend should be as follows:

Store API:
```
POST _integration/store/$instance_name
```
The $instance_name represents the specific name the integration was instanciated with - for example an Nginx Integration can be a template for multiple Nginx instances
each representing different domain / aspect such as geographic.

_For example the next observability integration:_

```json
  ...
  "collection":[
    {
      "logs": [{
        "info": "access logs",
        "input_type":"logfile",  
        "dataset":"nginx.access", <<- subject to user changes
        "namespace": "prod",<<- subject to user changes
        "labels" :["nginx","access"],
        "schema": "file:///.../schema/logs/access.json"
      },
    ...
```

During the UX interaction with the user, user can update data-stream details shown here:
![Screenshot 2023-03-01 at 11 28 00 AM](https://user-images.githubusercontent.com/48943349/222274241-f8689084-5ff5-432f-bc06-83546ac255ec.png)

If the user keeps all the original data-stream naming convention (namespace and domain) the next phase would be the validation of the integration prior to loading all the assets.

### Data-Stream naming update
In case the user wants to update the data-stream naming details - the next screens will be presented:

![Screenshot 2023-03-01 at 11 28 39 AM](https://user-images.githubusercontent.com/48943349/222274246-d1be30a6-32e9-4b91-b60a-1fc0df2fca02.png)

Selection of the naming convention may also display available existing data-streams that are selectable if the user wants
to choose from available ones.

#### Validate Integration

Once the Integration instance was stored in the integration `store` index, it will have a `loading` status as displayed in the
first image.

Next the integration instance will undergo a validation phase in which
 - assets will be validated with the schema to match fields to the mapping
 - assets that depend on index-pattern will expect the index to exist
 - datasource will be validated to verify connection is accessible
 - mapping templates are verified to exist

[...]

PR #1464, opened 2023-03-09
Implement Validator class for Integrations

Swiddis:
### Description
In preparation for implementing the integration validation, a validator that can ingest and process a schema needs to be implemented. This PR implements the `Validator` class that can be constructed from any of the current System Schema objects.

The tests for `Validator` also implement automatically checking sample json against their schemas at the unit test level, which will make API development more robust moving forward.

Future work:
- Endpoint for Integration Validation (issue pending)
- Make the class more robust with more schema support (issue pending)

Checklist before updating from Draft:
- [x] TODO in `validation/IntegrationComponent.kt`; I'm not sure what the best way to safely use relative paths in Kotlin is, open to suggestions.
  - Since there are unit tests that make sure this is set up right, it can also be deferred to properly resolving later. Marking as done.
- [x] Usage documentation

### Issues Resolved
Part of work on #1443 

### Check List
- [X] New functionality includes testing.
  - [X] All tests pass, including unit test, integration test and doctest
  - [x] New functionality has javadoc added
- [X] Commits are signed per the DCO using --signoff


src/main/kotlin/org/opensearch/observability/validation/IntegrationComponent.kt
```diff
@@ -0,0 +1,15 @@
+package org.opensearch.observability.validation
+
+private const val ROOT = "../../../schema/system/"
+
+enum class IntegrationComponent(val resourcePath: String) {
```
YANG-DB, line 5: these validation should be declared under the system catalog context (package ?)

src/main/kotlin/org/opensearch/observability/validation/Validator.kt
```diff
@@ -0,0 +1,56 @@
...
+            log.fatal("could not find schema '${schemaFile.path}' for component '$component'")
+            throw FileNotFoundException("could not find schema '${schemaFile.path}'")
+        }
+        val schemaSource = PathSchemaSource(schemaFile.toPath())
+        return medeiaApi.loadSchema(schemaSource)
+    }
+
+    fun validate(json: String): Result<XContentParser> {
+        return try {
+            val mapper = jacksonObjectMapper()
```
YANG-DB: can we cache this mapper for future use ?

src/test/kotlin/org/opensearch/observability/validation/ValidatorTests.kt
```diff
@@ -0,0 +1,89 @@
...
+    fun testConfigInvalidJson() {
+        val config = "{"
+        val validator = Validator(IntegrationComponent.INTEGRATION)
+        assertThrows<JsonParseException> {
+            validator.validate(config).getOrThrow()
+        }
+    }
+
+    @Test
+    fun testValidatorSampleIntegrations() {
```
YANG-DB, line 53: if are you expecting valid response here - the name should reflect this
also separate each schema validation to a different test
add per schema test different invalid validation tests with different cases

YANG-DB (changes requested):

src/test/kotlin/org/opensearch/observability/validation/ValidatorTests.kt
Swiddis: Split up the tests across methods, but I'm wondering how thorough we should be with per-schema checks. At some point that's just testing the validation library instead of our code. The main things to check here are that the paths are configured correctly, and that it's actually calling the validator, right?

YANG-DB (dismissed):

build.gradle
```diff
@@ -200,6 +200,11 @@ dependencies {
     implementation group: 'com.github.wnameless.json', name: 'json-base', version: '2.2.1'
     implementation "com.fasterxml.jackson.core:jackson-databind:${jackson_version}"
     implementation "com.fasterxml.jackson.core:jackson-annotations:${jackson_version}"
+    // Forced dependency from json-schema-validator, we add nop to deactivate it
+    implementation group: 'org.slf4j', name: 'slf4j-api', version: '2.0.6'
+    implementation group: 'org.slf4j', name: 'slf4j-nop', version: '2.0.6'
```
ps48, line 205: Is there some indirect dependency on `slfj` cause we usually use `log4j`

Swiddis: Per the comment, it's a forced dependency from `json-schema-validator`.

Swiddis:
Depends on #1482

YANG-DB (approved):

Swiddis:
Moved to #1486

PR #1531, opened 2023-05-10
bumb bwc version to 2.8

derek-ho:
### Description
Fix build failures related to outdated bwc version

### Issues Resolved

### Check List

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -319,7 +319,7 @@ testClusters.integTest {
     setting 'path.repo', repo.absolutePath
 }
 
-String baseVersion = "2.7.0"
+String baseVersion = "2.8.0"
 String bwcVersion = baseVersion + ".0"
 String baseName = "obsBwcCluster"
 String bwcFilePath = "src/test/resources/bwc/"


Swiddis (approved):

YANG-DB (approved):

build.gradle
```diff
@@ -319,7 +319,7 @@ testClusters.integTest {
     setting 'path.repo', repo.absolutePath
 }
 
-String baseVersion = "2.7.0"
```
vamsi-amazon, line 322: can we use opensearch_version instead of changing it after every release. Is it because the version bump is delayed?

Swiddis: The [redirect api](https://oss.sonatype.org/nexus-restlet1x-plugin/default/docs/path__artifact_maven_redirect.html) supports `LATEST` and `RELEASE` as alternative tags. `LATEST` defaults to 3.0 right now, but `RELEASE` seems to be unused, is there anything we can do to make that track the latest `2.x` snapshot?

I guess it would still be flaky because the versions might not all update at the same time, but it would also be self-fixing if we can set up the system for it.

PR #1620, opened 2023-09-12
Bump BWC version to 2.11

derek-ho:
### Description
Bump BWC version to 2.11
Relies on #1601 

### Issues Resolved

### Check List

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -303,7 +303,7 @@ testClusters.integTest {
     setting 'path.repo', repo.absolutePath
 }
 
-String baseVersion = "2.10.0"
+String baseVersion = "2.11.0"
 String bwcVersion = baseVersion + ".0"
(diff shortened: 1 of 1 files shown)


joshuali925 (approved):
did we miss this in auto version bump config?

derek-ho:
> did we miss this in auto version bump config?

I don't think there is a workflow to bump the BWC version in main to the newest one, only to bump the version on 2.x to the latest one, is there something I'm not aware of?

joshuali925:
> I don't think there is a workflow to bump the BWC version in main to the newest one, only to bump the version on 2.x to the latest one, is there something I'm not aware of?

i think you are right. i was thinking if it would require a manual PR after every release to update this in main, probably auto version bump should also cover it. not sure what other plugins are doing

ps48 (approved):

Issue #1883, opened 2024-11-11
[BUG] Observability plugin calling createIndex from onNodeStarted method in 2.x branches


shwetathareja:
**What is the bug?**

Observability plugin is implementing ClusterPlugin interface 

https://github.com/opensearch-project/observability/blob/ce58ffa7cbbfed2b34889f7cd5a7fdc32c200647/src/main/kotlin/org/opensearch/observability/ObservabilityPlugin.kt#L48
and onNodeStarted it calls

https://github.com/opensearch-project/observability/blob/ce58ffa7cbbfed2b34889f7cd5a7fdc32c200647/src/main/kotlin/org/opensearch/observability/ObservabilityPlugin.kt#L96

which calls createIndex
https://github.com/opensearch-project/observability/blob/ce58ffa7cbbfed2b34889f7cd5a7fdc32c200647/src/main/kotlin/org/opensearch/observability/index/ObservabilityIndex.kt#L89-L91

ClusterPlugin - onNodeStarted method is called from Node Bootstrap [(Node.java](https://github.com/opensearch-project/OpenSearch/blob/e68838819710d7040cf2b591590285f1b86f0da0/server/src/main/java/org/opensearch/node/Node.java#L1779)) and it causing UnhandledException in the main thread

        pluginsService.filterPlugins(ClusterPlugin.class).forEach(plugin -> plugin.onNodeStarted(clusterService.localNode()));

Node is not initialized. Cluster is not formed, why is it trying to createIndex from here?
The main branch doesn't have same code as well? why is there divergence?

```
OpenSearchTimeoutException[java.util.concurrent.TimeoutException: Timeout waiting for task.]; nested: TimeoutException[Timeout waiting for task.];
Likely root cause: java.util.concurrent.TimeoutException: Timeout waiting for task.
        at org.opensearch.common.util.concurrent.BaseFuture$Sync.get(BaseFuture.java:257)
        at org.opensearch.common.util.concurrent.BaseFuture.get(BaseFuture.java:82)
        at org.opensearch.common.util.concurrent.FutureUtils.get(FutureUtils.java:94)
        at org.opensearch.action.support.AdapterActionFuture.actionGet(AdapterActionFuture.java:79)
        at org.opensearch.action.support.AdapterActionFuture.actionGet(AdapterActionFuture.java:68)
        at org.opensearch.observability.index.ObservabilityIndex.createIndex(ObservabilityIndex.kt:108)
[...]

PR #1885, opened 2024-12-03
fix bug Observability plugin calling createIndex from onNodeStarted m…

YANG-DB:
### Description
BUG - fix bug Observability plugin calling createIndex from onNodeStarted method before cluster state is ready (waits for the cluster state to be ready)

```java
   override fun afterStart() {
           ....
           // wait for the cluster here until it will finish managed node election
            while (clusterService.state().blocks().hasGlobalBlockWithStatus(RestStatus.SERVICE_UNAVAILABLE)) {
                log.info("Wait for cluster to be available ...");
                TimeUnit.SECONDS.sleep(1);
            }
 
```

To introduce additional start command for all CI build which rely on the OpnSearch Docker image, as Github rollout the deprecation of the Node 16 on all it's CI-runner, as the result all existing Github which rely on the old version of Node.JS (Ex: actions/checkout@v3) failed due to the following errors:

```
/__e/node20/bin/node: /lib64/libm.so.6: version `GLIBC_2.27' not found (required by /__e/node20/bin/node)
/__e/node20/bin/node: /lib64/libc.so.6: version `GLIBC_2.28' not found (required by /__e/node20/bin/node)
```
image

The issue has been discovered over https://github.com/opensearch-project/opensearch-build/issues/5178 and fix has been verified and  applied on multiple OpenSearch plugin, such as: 
 - https://github.com/opensearch-project/ml-commons/pull/3223
 - https://github.com/opensearch-project/asynchronous-search/pull/676

### Related Issues
- #1883
- Partially resolves: https://github.com/opensearch-project/sql/issues/3168

### Check List
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Commits are signed per the DCO using `--signoff`.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/observability/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

### DIFF
--- a/.github/workflows/opensearch-observability-test-and-build-workflow.yml
+++ b/.github/workflows/opensearch-observability-test-and-build-workflow.yml
@@ -20,20 +20,23 @@ jobs:
       # using the same image which is used by opensearch-build team to build the OpenSearch Distribution
       # this image tag is subject to change as more dependencies and updates will arrive over time
       image: ${{ needs.Get-CI-Image-Tag.outputs.ci-image-version-linux }}
-      # need to switch to root so that github actions can install runner binary on container without permission issues.
-      options: --user root
+      options: ${{ needs.Get-CI-Image-Tag.outputs.ci-image-start-options }}
 
     env:
       ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: true
 
     steps:
-      - uses: actions/checkout@v1
+      - name: Run start commands
+        run: ${{ needs.Get-CI-Image-Tag.outputs.ci-image-start-command }}
 
+      - uses: actions/checkout@v4
+        
       - name: Set up JDK ${{ matrix.java }}
-        uses: actions/setup-java@v1
+        uses: actions/setup-java@v4
         with:
           java-version: ${{ matrix.java }}
-
+          distribution: 'temurin'
+      
       - name: Run Backwards Compatibility Tests
         run: |
           echo "Running backwards compatibility tests ..."
@@ -73,11 +76,12 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v4
 
       - name: Set up JDK ${{ matrix.java }}
-        uses: actions/setup-java@v1
+        uses: actions/setup-java@v4
         with:
+          distribution: 'temurin'
           java-version: ${{ matrix.java }}
 
       - name: Build with Gradle
--- a/.github/workflows/opensearch-observability-test-and-build-workflow.yml
+++ b/.github/workflows/opensearch-observability-test-and-build-workflow.yml
@@ -11,6 +11,8 @@ jobs:
   build-linux:
     needs: Get-CI-Image-Tag
     strategy:
+      # Run all jobs
+      fail-fast: false
       matrix:
         java: [11, 17, 21]
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


src/main/kotlin/org/opensearch/observability/index/ObservabilityIndex.kt
```diff
@@ -96,32 +96,41 @@ internal object ObservabilityIndex : LifecycleListener() {
...
-                }
-            } catch (exception: ResourceAlreadyExistsException) {
-                log.warn("message: ${exception.message}")
-            } catch (exception: Exception) {
-                if (exception.cause !is ResourceAlreadyExistsException) {
-                    throw exception
+        try {
+            // wait for the cluster here until it will finish managed node election
+            while (clusterService.state().blocks().hasGlobalBlockWithStatus(RestStatus.SERVICE_UNAVAILABLE)) {
+                log.info("Wait for cluster to be available ...")
```
ps48, line 102: Thanks for the fix! 
I had the same question that was raised by Shwetha in the original issue:
```
 The main branch doesn't have same code as well? why is there divergence?
```
Do we know why was this only needed for 2.x and what's the path ahead to converge these branches?

YANG-DB: @ps48   I'm not sure why this was removed in the main version
not really involved recently with this repository....

.github/workflows/opensearch-observability-test-and-build-workflow.yml
```diff
@@ -11,27 +11,32 @@ jobs:
...
     container:
       # using the same image which is used by opensearch-build team to build the OpenSearch Distribution
       # this image tag is subject to change as more dependencies and updates will arrive over time
       image: ${{ needs.Get-CI-Image-Tag.outputs.ci-image-version-linux }}
-      # need to switch to root so that github actions can install runner binary on container without permission issues.
-      options: --user root
+      options: ${{ needs.Get-CI-Image-Tag.outputs.ci-image-start-options }}
 
     env:
       ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: true
```
peterzhuamazon, line 26: Thanks can be removed as it is deprecated by github.

.github/workflows/opensearch-observability-test-and-build-workflow.yml
```diff
@@ -56,7 +61,7 @@ jobs:
           cp -r ./build/distributions/*.zip opensearch-observability-builds/
 
       - name: Upload Artifacts
-        uses: actions/upload-artifact@v1
+        uses: actions/upload-artifact@v3-node20
```
peterzhuamazon, line 64: Use v4, as v3 has been deprecated by github.

.github/workflows/opensearch-observability-test-and-build-workflow.yml
```diff
@@ -88,7 +94,7 @@ jobs:
...
```
peterzhuamazon, line 97: Same as above.

ps48:
There was breaking change in the `@v4 artifact upload` workflow which is leading to:
```
Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run
```

Can you please update the uploaded artifact name to be generated from the matrix used in the workflow. Here's an example of similar issue fixed in the dashboards-observability: https://github.com/opensearch-project/dashboards-observability/pull/2259/files

ps48 (approved):

